PR #62, opened 2015-06-13
Fix no right-click animation on client when opening crate

TheJulianJES:
Basicly return true on the client


TheJulianJES:
(If you don't want to merge it than it would be good if you can fix that for the official release because this can cause client-side dupes)

BluSunrize:
It's already fixed.

TheJulianJES:
Okay :)

PR #75, opened 2015-06-24
OC wires support

marcin212:
I would need these changes to the API so that I could create an addon to IE which would add OpenComputer wires.


BluSunrize:
Explanation as to why WireType is to be replaced with Connection?

marcin212:
I need a TE on the other side to detach OCNode

marcin212:
It is the easiest way, but You can add to API:
```
onConnectCable(Connection con)
onDisconnectCable(Connection con)
``` 
or
```
onConnectCable(IImmersiveConnectable otherside)
onDisconnectCable(IImmersiveConnectable otherside)
```

BluSunrize:
Approved. Will be in the next build.

PR #93, opened 2015-06-25
Update ru_RU.lang

TheVizzy:


BluSunrize:
There's already a russian translation. Also, as a rule of thumb: You never, ever edit stuff liek credits files. That's incredibly rude.

TheVizzy:
no problem

PR #97, opened 2015-06-25
Added Telecommunication connector and wire

marcin212:
c69b75803de03d9bfc66597164ed45ce


BluSunrize:
Woh woh woh
Not like this. Make it an addon.
I only provided API changes for this, not native support.

marcin212:
The point is that the API is... not flexible enough.
Example: I spent eight hours figuring out ways around some limitations, but gave up on cable support. I can't add a networking cable for OpenComputers in a simple, meaningful and actually useful way, because of the limitations you put on the Wire Cutters.

BluSunrize:
Well IE also wasn't ever designed with this in mind. I'll add some API fixes to my ToDo, but native OC support is kinda past the point

alucard87pl:
That was kinda the point, though. Since the API did not provide the hooks to add these externally in an easy way, a direct pull request adding the needed features seems the next logical step.

BluSunrize:
That is why I'm planning those hooks >_>

PR #125, opened 2015-06-27
MT Support Cleanup

Yulife:
**What has been changed:**
 - Output before Input as Minetweaker does it like that as well
 - Simplify some names of the ZenMethods because those were kind of complicated and now they're easier for the user to understand


BluSunrize:
Closed. Make it better. x3

PR #126, opened 2015-06-27
Correct Alignment of MT Support

Yulife:
**What has been changed:**
 - Output before Input as Minetweaker does it like that as well
 - Fluid before Input as Blu wants it like that ,_,
 - Simplify some names of the ZenMethods because those were kind of complicated and now they're easier for the user to understand
 - Now with 100% less API conflicts


BluSunrize:
Much better.

Yulife:
:heart_eyes:

PR #142, opened 2015-06-28
Lang Update 

Udosa:
It's my version of russian translate , i used russian dictionary and knowledge of russian language


Adaptivity:
717811c1314ea209f3ed33b965dfb47f
Go out please. If you want use this crap as translation, use a resourcepack. You know nothing about translation as well as your learnings of both Russian and English languages are bad.
Firstly, you wrote every word with capital letter - I hate you for this most of all. Secondly, you just copied-pasted all entries from google translator.
Thirdly, you already made a mistake at third line. Fourthly, you have no clue what are you translating (For example, you translated *Wooden Post* as *Wooden messages*, wtf!? And you saying that your translation is better? Log off from Github now). And fifthly, Beevolver can be translated. It's *Bee* plus *Revolver*, if you won't translate it (or can't), it doesn't mean that it can't be translated.
@BluSunrize sorry if my words too aggressive here. I tried to be patient as much as possible.

BluSunrize:
Beevolver shoudl not be translated however, since it's a unique name. Change that back the next time you make an update, @Adaptivity? =3

Udosa:
@Adaptivity if it's you translated TiC then i all understand TT_TT

Adaptivity:
I don't see the reason of not translating this. I only see the reason if the name can't be translated or translation will be really derpy. For example, Chinese translators translating everything (In all mods, not just yours). And yeah, they already [translated beevolver](https://github.com/3TUSK/IE-Simplified-Chinese-Localization/blob/master/src/main/resources/assets/immersiveengineering/lang/zh_CN.lang#L152). PR should come when they finish translation I guess. But if you are not agree with me, ok I can revert it, not a problem.
@Udosa I'm translated 99% of mods which have Russian language support. You have no rights to judge my translations. I spent seconds, minutes, hours, days, weeks, months, years working on these translation for free. If you think your translation is better, use resourcepack. I see no point in further discussion of my translations.

3TUSK:
@Adaptivity Agreeing with your opinion of "PR should come when whey finish translation", [I am port the IE translation to somewhere else for convenience.] (https://github.com/3TUSK/Temp_LocalizationFile/tree/master/ImmersiveEngineering)
Also here is another thing. Sometimes we (there are a lot of people working on Chinese localization for minecraft mods) are unable to translate everything, because it is a common thing for Chinese translators that there are a lot of stuff which requires translators' *special imaginary* to do with them. A good example is "Vazbee" from Magic bees, making up from Vazkii+Bee obviously, but it is hard to be translated into Chinese unfortunately. The reason is that we expect the Chinese name of every Forestry bee to have only two characters , but it seems that "Vazkii" has 3 syllables. (If there is something wrong please point out.)
Under such a situation, we either find homophones in Chinese or just keep English (or use acronym, rarely) here, mostly to show the respect to original context, although it may cause the loose of original meaning.

The same situation as Immersive Engineering. My friends and I have noticed some items' names with their special origin, and these names are under further discussion currently. To be more accurate, I get some questions to ask @BluSunrize , if you are happy to answer:
1.Does the "immersive" in "immersive engineering" have the idea that "generating a three-dimensional image which appears to surround the user" (from google), or just mean "sinking (soaking?) in something else"?
2.One of my friend claimed that [the oathkeeper and oblivion] (https://github.com/BluSunrize/ImmersiveEngineering/blob/master/src/main/resources/assets/immersiveengineering/lang/en_US.lang#L150) are from [Kingdom Hearts] (https://en.wikipedia.org/wiki/Kingdom_Hearts). Is he right?
Thanks for your time.

(Doesn't this comment look pointless?! Probably because I am sleepy while writing it. :( )

bakaxyf:
[Fenrir and Earthshaker] (https://github.com/BluSunrize/ImmersiveEngineering/blob/master/src/main/resources/assets/immersiveengineering/lang/en_US.lang#L153) are from Kingdom Hearts too.

BluSunrize:
Immersive means "soaking" in a way.
And yes, Oathkeeper, Oblivion, Earthshaker and Fenrir are all from Kingdom Hearts.
All the names for the Revolvers are "unique" names though, they all have a very specific meaning, that's why I thought it silly to translate them.

3TUSK:
Thanks a lot for answer first.
However, did people who cannot understand English play with IE, those untranslated names would be a loose because they would unable to catch the idea(s) beneath those names. As long as there is possibility, we won't give up translating names of Revolvers. Translation is aim to provide the same feeling under different language situation, isn't it?

BluSunrize:
They also, probably won't get to see most of those revolver names, since they are exclusive to a few people .-.
Feel free to translate Last Word and Twisted Phantom, since those are for patreons, but Beevolver, Earthshaker and Steamroller are unlikely to be seen "in the wild"

PR #154, opened 2015-06-30
Update .gitignore

xbony2:
This is so mac users like me can send pull requests without the .DS_Store file that are auto-generatored by OS X, since most people don't those. See: http://en.wikipedia.org/wiki/.DS_Store


BluSunrize:
I know nothing of Apple, so I'll just approve this. xD

PR #175, opened 2015-07-04
Allow addons to access and add categories to the manual

UnwrittenFun:
Adds a method to get access to the manual.

Adds a method to add a custom category to the manual.


BluSunrize:
That's a very good point. But I'll go for a slightly different implementation of it =P

Thanks for the input though =)

UnwrittenFun:
I though you might, I just did this to play around with adding to the manual and thought if you weren't planning on adding support at least you could pull mine, looking forward to your implementation :)

BluSunrize:
Check https://github.com/BluSunrize/ImmersiveEngineering/commit/c05b14ec02c8d14e9f65014d8d915fa10c9c908d =P

PR #177, opened 2015-07-04
add maven dependencies

cobra:
add maven dependencies
fix gradlew file permissions on unix systems

todo: handle dependencies without maven repositories (AquaTweaks, Minefactory Reloaded, Minetweaker). I don't know how you want to handle these, so I omitted them. One possible, but hacky, solution could be this: https://gist.github.com/matthewprenger/2772a516fae6069d6bb4


BluSunrize:
Nope. I don't do maven, I do my libraries manually. This pull won't be happening.

PR #186, opened 2015-07-05
Add ForgeMultipart compat

miuirussia:
Please, add ForgeMultipart support


AEnterprise:
can this even work? i know carpenters blocks don't work with these cause of the special rendering

miuirussia:
Tested and working)

XFactHD:
@AEnterprise it works (I tried it by simply adding the decorative blocks to FMPs multipart.cfg) but the covers (at least in this aproach) are unable to disconnect FMP cables like OCs networking cables.

BluSunrize:
Multiparts are a thing now anyway. I forgot abotu this one, and jsut added them when Damien bugged me xD

PR #194, opened 2015-07-06
getRandomMineral is actually random

marcin212:
before fix: http://ideone.com/KWhkBc
after fix: http://ideone.com/C2xmaQ

If "weight" is smaller then 0, then I can find only Iron, because Iron is first on mineral list.


BluSunrize:
Same should be achieveable with Math.abs() if I'm not mistaken?

marcin212:
yes

marcin212:
I changed it.

BluSunrize:
woops.

PR #200, opened 2015-07-08
Wrong translation

bakaxyf:


BluSunrize:
I can't read it so I'll just pull it ^^'

PR #212, opened 2015-07-12
Lightning bolt is visible now.

marcin212:


BluSunrize:
forgot about this one....

PR #215, opened 2015-07-13
Make gradlew executable

Kubuxu:
`gradlew` is not executable on linux systems.


BluSunrize:
I don't dev on linux. If you must, set up your own environment.

Kubuxu:
It is just a file permission change. Now every dev on linux has to do `chmod +x gradlew` to start gradle and it will conflict on any rebase and so on.

PR #216, opened 2015-07-13
Add config option for Thermoelectric Generator output

Kubuxu:
Default values can be too high in some cases where energy is important part of balance.
Thermoelectric gives free energy and it might require nerfs in some cases.
I've removed unboxing as it is not necessary (Java does it by itself, null check is enough).
Also added `out` dir to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@
 /copy
 /git
 /world
+/out
 options.txt
 eula.txt
 server.properties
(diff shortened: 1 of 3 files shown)


BluSunrize:
I'll need to do this manually, because my local repo is ahead. I added the modifier for the thermoelectric gen already upon suggestion from another user I just didn't push it yet. Will add the IDEA stuff though.

Kubuxu:
Ok, thanks.

PR #253, opened 2015-08-01
Improved build.gradle

Techokami:
Several dependencies are now automagically obtained through maven repos,
meaning you only need 3 deobf jars in your libs folder: AquaTweaks,
MineFactoryReloaded, and MineTweaker. This makes compiling the mod a lot
easier, as you don't need to run around the internet grabbing deobf jars
for a ton of mods this depends on, based on errors in the build log.
Though ideally, you'd break out version numbers into variables stored in
a build.properties file...

There was another change, but I scrubbed it out.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,21 @@ buildscript {
     }
 }
 
+repositories {
+    maven { // The repo from which to get ChickenBones' stuff, like CCLib and NEI
+        name 'CB Maven FS'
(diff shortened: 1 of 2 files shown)


marcin212:
Look at https://github.com/BluSunrize/ImmersiveEngineering/pull/177

Techokami:
ooh, did not know that.

BluSunrize:
Yep.

PR #260, opened 2015-08-03
Fix slag output of Arc Furnace

cobra:
Auto-outputting slag onto conveyor belts or other sided inventories didn't work.


BluSunrize:
Oh! Well spotted, thanks :D

PR #291, opened 2015-08-07
Fix HV Wire Relays, close #288

### LINKED ISSUE #288: Cant connect to HV wire relays
whenever i try to connect to an HV wire relay, i get the message "the connection is obstructed" even when it obviously isn't. also, trying to connect FROM a relay works

cobra:
I simply added a missing method override with the values calculated the same way as for the other connector types. Maybe there could be a more elegant solution, but I know you hate constants :p


BluSunrize:
Thank you kindly =P

PR #377, opened 2015-09-01
Fix not being able to place blocks if stood over the place the block …

UnwrittenFun:
…should go.

As shown: http://unwrittenfun.uk/sharex/files/2015-09-01_03-45-54.mp4


BluSunrize:
I don't even know why this is a thing, but Mojang. xD

UnwrittenFun:
It because the item block checks if you are colliding with the collision box  by getting it from that method, but the block isn't in the world at that point so getting the meta is always going to return 0, hence giving you the fence collision box

BluSunrize:
Ooooooh, That makes sense, actually xD

Issue #346, opened 2015-08-20
Crusher sound/animation issue


theunkn0wn1:
@BluSunrize Managed to repo #63....
Rather than necro that old issue I am creating a new issue thread.

Issue: Crusher continues to act in its 'working' state after power is terminated
=========================

If the crusher is processing an item when it looses power, it continues to act in its 'working' state, animation, sounds, and particle affects

Steps to reproduce:
-----------------
[1] Build Crusher MB correctly
[2] supply Crusher with power
[3] supply crusher with items to process such as iron ore)
[4] cut power to Crusher while it is still processing the item
[5] observe

Version data:
-----------------
[IE version 0.5.3](https://github.com/BluSunrize/ImmersiveEngineering/commit/9200be056b205f7680314b9a0c9504c0cfd505f3)
FTB Infinity Version 1.9.0 custom botania

cobra:
I already looked a bit into it, here's what I found so far:
- without power, the client doesn't ever get updated when the "active" field is changed after the crusher finishes processing.
- adding some power connection masks this because RF causes updates
- removing `if(update)` or replacing it with `if(true)` in https://github.com/BluSunrize/ImmersiveEngineering/blob/master/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityCrusher.java#L215 fixes this as well

I could not yet find when or where something is missing an update trigger.

malte0811:
I believe that I have fixed this in the last two commits of https://github.com/malte0811/ImmersiveEngineering/tree/fix-1. The first one fixes it running (animation) when the power buffer is empty but the item is not completly processed, the second one fixes it running after the item has finished processing if no power is supplied. The client update containing the data for the particles was being sent after processing rather than before, which set `active` to true and `process` to 1 client sided

cobra:
I didn't yet look into your second commit for the sync problems, but I took a different route for stopping the animation and sound. I checked whether there was any energy consumed in https://github.com/BluSunrize/ImmersiveEngineering/blob/master/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityCrusher.java#L139 and changed the state accordingly without adding yet another field to check.

I have to test the whole a bit later, especially your update changes.

cobra:
[...]

PR #380, opened 2015-09-01
Fix crusher problems

cobra:
fixes all the currently known issues of the crusher, see #346.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityCrusher.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityCrusher.java
@@ -144,14 +144,6 @@ public class TileEntityCrusher extends TileEntityMultiblockPart implements IEner
 					{
 						ItemStack inputStack = inputs.get(0);
 
-						if(inputStack!=null)
-						{
-							Block b = Block.getBlockFromItem(inputStack.getItem());
-							int id = (b!=null&&b!=Blocks.air)?Block.getIdFromBlock(b): Item.getIdFromItem(inputStack.getItem());
-							int meta = inputStack.getItemDamage()+((b!=null&&b!=Blocks.air)?0:16);
-							worldObj.addBlockEvent(xCoord,yCoord,zCoord, this.getBlockType(), id,meta);
-						}
-
 						CrusherRecipe recipe = CrusherRecipe.findRecipe(inputStack);
 						if(recipe!=null)
 						{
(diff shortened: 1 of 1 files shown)


BluSunrize:
:O
Thank you. I was fearing that I'd have to assemble this one from the crusher report and the forks <3

malte0811:
@cobra Now that i am seeing your code for the problem i solved with `hasEnergy`, i see a minor problem in it(haven't tested it yet, but i am pretty sure): if the crusher gets powered after having run out of energy, it finishes using the energy for whatever it was processing (count process down to 0) and then starts processing the same item again since active is false. That was probably why i used a new variable. Also your code syncs the nbt every tick when there is no power, the `else` should, if you ignore the other problem, be an `else if (active)`

cobra:
@malte0811 I even asked for comments before making this PR :/
Yep, you're right. But this is easily fixed by checking `active` first before setting and updating it: https://github.com/cobra/ImmersiveEngineering/commit/1ef50b386bcc3b7509329cff9fc7284609503be2

BluSunrize:
I implemented stuff that @malte0811 suggested with this
https://github.com/BluSunrize/ImmersiveEngineering/commit/01e384ad751770d90f6d391e95fffb27b90efc45

cobra:
Yep, I just saw that. But you would still need to pull in this commit since that part is still broken, right? I'm just looking at your changes.

malte0811:
@BluSunrize The issue is still there, you should probably revert a8a8c2ca1d7a412a2adbf704aaee6d92d631c7c9. You are still setting active to false when there is no power, causing the crusher to re-start processing the item it was processing once it has power and has finished consuming the energy for the first run. If whatever power supply the crusher is connected to does not produce rf every tick, the crusher will never process anything
@cobra Sorry, i did not see that problem when i had a quick look at your code before

BluSunrize:
Gah, seriously guys, you need to get on the same level here xD
One tells me this, the other tells me that, I'm just bouncing between commits here.
If you feel it's a necessary adjustment to the code, shoot me a PR detailing the changes, but don't expect me to create code on the basis f a discussion that I haven't been following xD

cobra:
You'll get a PR since as of now, your master branch is broken and this would be the easiest fix.

Technically, this bug (crusher appears to be running while doing nothing) is now double-dead since you've got both solutions for it, which won't conflict with each other.

BluSunrize:
hahah, alright, glad it's done then xD

PR #389, opened 2015-09-03
fix crusher updates when out of energy

cobra:
Only change `active` state of the crusher when it needs to be changed, avoiding unnecessary updates. (`else if(active)` part)
Also resume processing instead of just consuming energy and starting processing for the same item again. (`if(!active)` part)


malte0811:
Please don't pull this until 9pm german time, i think there is a hard to explain bug in there, i want to test that

malte0811:
I was wrong, I forgot that the changed code only runs if `process>0` is true. I personally don't see any problems with merging this now

cobra:
Yay. I tested with some rather weird patterns to make sure this is bug-free and couldn't find anything, so I was quite intrigued by your comment.

Issue #268, opened 2015-08-04
Arc furnace (technical block?) is available in the creative tab


planetguy32:
![2015-08-04_09 21 08](https://cloud.githubusercontent.com/assets/2831434/9066061/52c50190-3a8a-11e5-8d0b-fb34f4cc9c2d.png)

When placed, it's shaped like a half-slab and has no texture. It has a TileEntityArcFurnace, and when I punched it I got this crash:

java.lang.ArrayIndexOutOfBoundsException: -1
	at blusunrize.immersiveengineering.common.blocks.metal.TileEntityArcFurnace.getOriginalBlock(TileEntityArcFurnace.java:50)
	at blusunrize.immersiveengineering.common.blocks.metal.BlockMetalMultiblocks.breakBlock(BlockMetalMultiblocks.java:138)
	at net.minecraft.world.chunk.Chunk.func_150807_a(Chunk.java:658)
	at net.minecraft.world.World.setBlock(World.java:519)
	at net.minecraft.world.World.setBlockToAir(World.java:651)
	at net.minecraft.block.Block.removedByPlayer(Block.java:1670)
	at net.minecraft.block.Block.removedByPlayer(Block.java:1664)
	at net.minecraft.server.management.ItemInWorldManager.removeBlock(ItemInWorldManager.java:270)
	at net.minecraft.server.management.ItemInWorldManager.removeBlock(ItemInWorldManager.java:263)
	at net.minecraft.server.management.ItemInWorldManager.tryHarvestBlock(ItemInWorldManager.java:304)
	at net.minecraft.server.management.ItemInWorldManager.onBlockClicked(ItemInWorldManager.java:168)
	at net.minecraft.network.NetHandlerPlayServer.processPlayerDigging(NetHandlerPlayServer.java:523)
	at net.minecraft.network.play.client.C07PacketPlayerDigging.processPacket(C07PacketPlayerDigging.java:61)
	at net.minecraft.network.play.client.C07PacketPlayerDigging.processPacket(C07PacketPlayerDigging.java:94)
	at net.minecraft.network.NetworkManager.processReceivedPackets(NetworkManager.java:241)
	at net.minecraft.network.NetworkSystem.networkTick(NetworkSystem.java:182)
	at net.minecraft.server.MinecraftServer.updateTimeLightAndEntities(MinecraftServer.java:726)
	at net.minecraft.server.MinecraftServer.tick(MinecraftServer.java:614)
[...]

PR #394, opened 2015-09-04
remove silo/tank technical blocks from creative tab

cobra:
same problem as #268
you don't have a maximum line length, do you?


BluSunrize:
Well IE 0.6 isn't released yet, so this isn't really a public issue. I just didn't get aroudn to this yet.
But thanks anyway. xD

PR #399, opened 2015-09-05
fix Engineer's House double doors

cobra:
The right door was oriented with the hinges on the left and opening in
the middle of the double door instead of the right side. Minecraft
checks for a door left of the door it's placing, so changing the
placement order fixes this automatically.


BluSunrize:
GG Mojang, I guess. TY =P

Issue #392, opened 2015-09-04
Wire connector validation wacked up.


theunkn0wn1:
Issue: Wire validation is... wonky

when attempting to connect two wire connectors of the same type together, there are quite a few scenarios where the LOS is clear yet "the connection is obstructed"
I have attached several images below showing this.

![image](https://cloud.githubusercontent.com/assets/3110986/9677666/ab991210-5291-11e5-913f-8bd4202819e1.png)
Unable to connect the HV Connector to the relay
![image](https://cloud.githubusercontent.com/assets/3110986/9677677/d6ad4872-5291-11e5-92c4-6742c9ddb786.png)
![image](https://cloud.githubusercontent.com/assets/3110986/9677691/f3f1b260-5291-11e5-97d3-82c2ed9692e3.png)
That last one works from one direction but not the other

![image](https://cloud.githubusercontent.com/assets/3110986/9677726/51661da0-5292-11e5-80a3-8f407a297ca7.png)
I had to remove the wall mount to allow the MV connectors, whom have LoS with the wall mounts in place, for the connection to validate
![image](https://cloud.githubusercontent.com/assets/3110986/9677778/dbf0c650-5292-11e5-967e-a943e985229c.png)
between the breaker and the first relay, <1 meter distance

running latest IE build, 0.6.0_pre c584f5f6af907554a718841a7dea0a14f33eb054

BluSunrize:
Raytracing is a BITCH. Seriously, I have sunk so much time into it and it's still really wonky.
Problem is that I dunno how to properly fix it, and I've tried quite a few things, so thsi is low priority for now

theunkn0wn1:
Idk what you did as of late, but its been more wonky than normal. I reported this because it got worse, not better/stayed same. Before it was acceptable, but a little twitchy. Now, its just got that itchy nope.avi finger on hotkey.

Barhandar:
Which point of the connector block it's raytracing from?

theunkn0wn1:
as i understand its supposed to be the tip of the connector; the end opposite the mounting side. for relays its 'bottom' with a little offset. Its supposed to be the same point the cables render draws to/from
Did i get that right @BluSunrize?

BluSunrize:
You kinda did. Every IImmersiveConnectable can return its own specific raytrace offset. I usually specify these with a little leeway, to avoid colision inside the connector itself.
[...]

PR #411, opened 2015-09-11
Restart connection raytracing

cobra:
This should remove nearly all unexpected "connection obstructed" messages due to failed connection validation, see #392. The only exception are blocks with an irregular model like wall mounts - we can only check for bounding boxes, not visible models.

ToDo before or after merging: remove debug logging and threshold variable

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/BlockMetalDevices.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/BlockMetalDevices.java
@@ -417,7 +417,7 @@ public class BlockMetalDevices extends BlockIEBase implements blusunrize.aquatwe
 	{
 		if(world.getTileEntity(x, y, z) instanceof TileEntityConnectorLV)
 		{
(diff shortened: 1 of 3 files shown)


BluSunrize:
thank you <3

cobra:
Remember to remove the debug stuff (logging, adjustable threshold) before releasing 0.6. I left it in to allow easy testing without having to use the debugger each time.

BluSunrize:
Wait, if I remove the debug threshhold, what else do I take out? it's actively used in the conditions, isn't it?

cobra:
replace it by the value, there's no need to have that as a variable besides allowing you to change it during runtime for testing purposes. I didn't find any situation needing lower thresholds, but maybe someone else finds one where ±0.4 would be necessary to allow a valid connection.
Remove it like this: https://github.com/cobra/ImmersiveEngineering/commit/078e254eeed3108136b5deaefaa4ceae504db265

BluSunrize:
Ah, right kk =P

PR #412, opened 2015-09-11
Fixes for the energynet

malte0811:
fixed the `Comparable`-implementation of `ImemrsiveNetHandler.Connection`, as it was not "consistent with equals" (http://docs.oracle.com/javase/7/docs/api/java/lang/Comparable.html), this fixes a lot of weird behavior like the wire cutter behaving "weird" and the creation of "diode"/"one-way"-wires
fixed the limiter being reset exactly when it shouldn't (This fixes the HV-connector bug @cobra mentioned)
fixed wires not actually getting removed in `ImmersiveNetHandler.clearAllConnectionsFor(ChunkCoordinates, World, TargetingInfo)`, as they were being removed from a copy rather than the real set

There probably are more bugs in the energy net, but those are the ones I found and was able to fix.


BluSunrize:
Thank you <3

PR #422, opened 2015-09-13
Fix a small bug with electric lanterns

Guichaguri:
That's why the lantern was not saving the energy storage =P


BluSunrize:
I actually fixed that already, just didn't commit it. Thanks though xD

PR #424, opened 2015-09-15
Some small floodlight things.

mrammy:
Fixed a typo preventing the Floodlight render from loading. Added a localization for the floodlight.

I hope this helps you get the more important stuff done quicker. I am really enjoying your mod.

mrammy


BluSunrize:
How does this affect the render?

mrammy:
it only changes the name to make your object load

cobra:
@UnwrittenFun broke [this line](https://github.com/BluSunrize/ImmersiveEngineering/blob/master/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/BlockMetalDevices2.java#L43) again when resolving conflicts for #444. -.-
@BluSunrize you need to redo this change when continuing your work on the flood light. BTW: awesome model and idea. Will the direction affect the light (being more beam-like) or is this just for cosmetic purposes?

BluSunrize:
Thank for poiting that one out @cobra. And yes, I want it to be beam like. Sadly, Damien doesn't like my model, which is why I haven't worked further on it yet.

cobra:
That model is awesome, I just think the texture could be a bit brighter.

PR #431, opened 2015-09-17
only grow Hemp from the first block, close #430

### LINKED ISSUE #430: Hemp crop height bug
http://pastebin.com/3fYUfHGN < Mod List

http://i.imgur.com/1yLpiWp.png < Overview of issue.

http://i.imgur.com/TNFtF51.jpg < Showing every 10 blocks normal hemp crop growing for 2 blocks.

Hemp crops growing without limit when using project e's Harvest Goddess Band in player inventory, as long as player stays within range of the top of the crop it will continue to grow.

After 10 blocks of missing textures it will grow 2 normal textured crops and continue again with missing texture, this happens every 10 blocks.

How to reproduce: Equip the Band, Stand near the crop, Staying level with the top of the crop, Crop will continue to grow.

cobra:
Some mods don't check func_149852_a before calling func_149853_b,
causing invalid metadata and endless growth.


BluSunrize:
Thanks =3

PR #432, opened 2015-09-17
Fix #428 cables being removed from wrong world.

### LINKED ISSUE #428: Wire fry/render issue
Issue: Mv wires, when fried, do not always destroy/invalidate the connections.

When using Mv wires to draw off the power generated from a diesel generator, **intentionally**, I found a case where the wires do not properly fry from overload.

Possible steps below:
[1] Build diesel generator correctly
[2] place three Mv Wire connectors to the OUTPUT faces of the generator.
[3] wire all three together in series
[4] wire one of these connectors to a RF consumer. (in my instance a Trash Can)
[5] replace fried cable starting from consumer to provider (order specific!)

I am having dificulty understanding what exactly i have done, will be more clear [here](https://youtu.be/hBqWUtv-QN0) in a youtube video

.. IE Version 0.6.0_pre SHA 2fec12bbd2ec1a4981de46236e278bc4124c2857

marcin212:
Connections has been removed from first ticked dimension, which doesn't happen to be the right way to do things.


marcin212:
So, I am waiting for your pr.

PR #433, opened 2015-09-17
Fixes for the energynet, round 2

malte0811:
fixed some indirect connections not being used due to being considered equal to other connections without being so
fixed connections only being saved to nbt in one direction because they were considered equal to their reverse connection by `compareTo` and `equals`
fixed wires only burning in the "lowest" loaded dimension (this fix was found independently by @marcin212 as well)
this does create a merge conflict, for further information, see [here](https://help.github.com/articles/resolving-a-merge-conflict-from-the-command-line/)

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/api/energy/ImmersiveNetHandler.java
+++ b/src/main/java/blusunrize/immersiveengineering/api/energy/ImmersiveNetHandler.java
@@ -7,7 +7,6 @@ import static blusunrize.immersiveengineering.api.ApiUtils.toIIC;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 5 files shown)


BluSunrize:
question: Why does it create a merge conflict? Why not work against the latest commit?

malte0811:
I started working on this when @marcin212 reported the indirect connection bug, so when the commit i was working against was the latest one, and forgot to pull from your master branch in between.

marcin212:
I looked at the conflict and is easy to repair.

BluSunrize:
can't you just update to latest?

malte0811:
Closing this since @marcin212 made a copy without the conflict.

PR #438, opened 2015-09-19
remove only arms of the wooden post being broken

cobra:
This commit adds a check to make sure only the arms of the wooden post being broken are being removed. Without this check all arms around the post's top block would get broken, no matter which post they were attached to.
Example of a situation where this happend: 
Breaking the right post broke the arm of the left post, dropping the HV relay.

Using `fd.ordinal()` is somewhat hacky, but Forge doesn't offer a method to convert a ForgeDirection back to a 'side' int. It's the same way `type` is set when adding the arms, so it shouldn't matter and I didn't want to rewrite the whole code touching this field.


BluSunrize:
Heh, never noticed that one.

PR #441, opened 2015-09-20
various bugfixes

### LINKED ISSUE #374: [0.5.4] Negative RF flowing out of too long transmission line due to transmission losses
I was testing whether power lines can go across unloaded chunks and built a quite long line (~400 blocks). On the end I was getting negative RF accumulating in the capacitor I attached (which I assume should not happen). I did not investigate further what happens to negatively-charged capacitor (I was planning to set loss to 0 in config anyway, which of course removed the issue).

IE 0.5.4, inside Horizons 2 Daybreaker FTB pack 1.0.2 (unmodified, except for one mod - Computronics, removed)

### LINKED ISSUE #439: Mirrored Generator redstone control is not solid.
When the diesel generator is mirrored it does not make the block that is the controller on the opposite side solid so you can't place a lever on it like you can with the default side. Also the block where the redstone control was on the original side doesn't stop being solid after the switch. I am not sure if I have described the issue well enough so here is picture of generator that has been mirrored with a lever sitting where the redstone control is in the generator's normal state.

image

Step to reproduce:
Build generator multiblock in any orientation. (North/South/East/West)
Turn the multiblock into machine using hammer.
Then mirror the generator by click on the front.
Try placing a torch on the redstone control.

cobra:
The commit messages should be enough - tl;dr:
fix #439 (mirrored Diesel Generator control input not solid).
Don't drop some stuff when doTileDrops is set to false.
Stop losing more power than is available, fixes #374.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/api/energy/ImmersiveNetHandler.java
+++ b/src/main/java/blusunrize/immersiveengineering/api/energy/ImmersiveNetHandler.java
@@ -498,7 +498,7 @@ public class ImmersiveNetHandler
 			float f = 0;
 			for(Connection c : subConnections)
 				f += (c.length/(float)c.cableType.getMaxLength())*c.cableType.getLossRatio();
-			return f;
+			return Math.min(f,1);
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


BluSunrize:
thanks =3

PR #451, opened 2015-09-24
small fix for burning wires

### LINKED ISSUE #428: Wire fry/render issue

malte0811:
fixes #428
In singleplayer it is not necessary to use packets for changes in the `ImmersiveNetHandler` since the instance is a static variable that both server and client can access. Further explanation can be found in my comments on #428.


BluSunrize:
Thanks =)

PR #458, opened 2015-09-26
avoid crashing without CoFHLib/Core installed

### LINKED ISSUE #456: CoFH Hammer API missing
Here we go again, another missing class when launching MC with IE without a third-party mod.

Crashes on startup with 
````
java.lang.NoClassDefFoundError: blusunrize/immersiveengineering/common/items/ItemIETool
	at blusunrize.immersiveengineering.ImmersiveEngineering.preInit(ImmersiveEngineering.java:65) ~[ImmersiveEngineering.class:?]
````
Built with CoFHLib-[1.7.10]1.0.3-175-dev.jar, launched in MultiMC with only CodeChickenCore and NEI, deliberately missing any CoFH mods. Obviously, not a big surprise.
Most people will have CoFHCore, so they won't run into this problem, but I think this should be resolved - I just don't know which way you want to do this. I'd add the item package into the sources just like you did with the energy package. Also I don't have that (internal, not file name) version 1.0.10 (oh CoFH, you and your weird versioning), so maybe just update to the latest one, shouldn't change anything used in IE.

IE version: 0fbca858e93005bc304e171d902488e6279007ca

cobra:
closes #456

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/items/ItemIETool.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/items/ItemIETool.java
@@ -33,7 +33,7 @@ import com.google.common.collect.ImmutableSet;
 
 import cpw.mods.fml.common.Optional;
 
-@Optional.Interface(iface = "cofh.api.item.IToolHammer", modid = "CoFHAPI")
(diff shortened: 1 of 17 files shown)


BluSunrize:
Thanks =3

PR #460, opened 2015-09-26
Add a config option for EE3 compat

malte0811:
This is mostly for cases in which DynEMC breaks for no apparent reason. It allows changing between pre, post and not-at-all registered values for IE items.
I experienced the disconnect described in Pahimar/Equivalent-Exchange-3#967 when trying to use 0.6-pre in a modpack on a dedicated server, this was fixed by switching to post-values. In a different setup pre registered values broke DynEMC resulting in only base items having an EMC value, this was also fixed by pos-values.


pahimar:
There are still outstanding issues with EE3's DynEMC system, so in the meantime using Post-Calculation values would address the gap in the meantime (I would recommend players not modders use this option) until I can get the system working properly in all cases again.

malte0811:
@pahimar I know that it is a month since you commented, but I don't really understand what you want me to do about this PR. Do you want @BluSunrize to merge it, do you want it to be not merged at all or do you want me to remove option 1 (post values)? (I am asking this now since the update is probably going to happen soon)

malte0811:
Closing this since compat for every mod can now be toggled.

PR #469, opened 2015-09-28
add lanterns to engineer's house, fix double door

cobra:
Doors automatically become double door when placed in the correct order,
but building a rotated house breaks this mechanic, so rotation needs to
be taken into account.
I thought I had fixed that in c182d9fb1e45c35196a7f8a23141925f8e96b533, but that was only one half. Vanilla is weird, having fancy double door checks but breaking them at the same time -.-
I also added two lanterns to the engineer's house. It seemed somewhat strange that the tech guy of the village is living in a completely unlit house. Also, torches are boring and everyone has them.


BluSunrize:
Sure, whatever, ty

PR #472, opened 2015-09-29
World save data fixes

cobra:
Makes loading world save data a bit more reliable, mainly countering missing wires. The commit messages and the issues have all the details.


BluSunrize:
Let's hope this actually fixes stuff .~.

cobra:
I did rather extensive testing on that in addition to the debugging to find what was causing these weird behaviours and couldn't see anything getting worse…

XFactHD:
Did you also test what happens when performing a debug crash without saving by pressing escape?

cobra:
This doesn't change anything on that end, saving works exactly the same as before. I changed some cases in which that data wasn't loaded, completely ignoring what was saved.

XFactHD:
Ok, nevermind.

PR #475, opened 2015-09-30
Reduce the number of World.getTileEntity calls, closes #467

### LINKED ISSUE #467: Too many getTileEntity() calls
Hi there!
I was reading through your code when something that caught my attention: you call World.getTileEntity() a lot, even when you have already called it once for the same world location and getTileEntity() is not the fastest function around.

For example, in TileEntityExcavator.updateEntity() you call it 5 times for the same world location

You really need just one call. 

The same thing goes for other function that compute and return a value that you need to use, like the master() method of your multiblocks tile entities

I'm just pointing this out because I really love to play with IE

malte0811:
The title basically says it all, #467 contains more information. The changes are partially by @cobra and partially by me.


BluSunrize:
want me to pull it anyway, even though it'S ugly? >_>

cobra:
what is ugly? *whistles* go for it.

BluSunrize:
50% of IE is your guys code by now .~.

PR #497, opened 2015-10-11
fix rendering for mirrored multiblocks

cobra:
These three multiblock structures weren't rendered correctly when mirrored and oriented in north/south direction (affecting 2 out of 8 possible orientation/mirrored states).

Technically, the same problem exists for the assembler and the refinery, but these two cannot (yet?) be rotated. The rest of the multiblocks works fine - yes, I've built them all 8 times.

Most obvious when looking at invisible blocks or clipping into the model, besides parts being in the wrong place or the model facing the wrong direction:

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/client/render/TileRenderArcFurnace.java
+++ b/src/main/java/blusunrize/immersiveengineering/client/render/TileRenderArcFurnace.java
@@ -41,7 +41,8 @@ public class TileRenderArcFurnace extends TileRenderIE
 		TileEntityArcFurnace arc = (TileEntityArcFurnace)tile;
 
 		translationMatrix.translate(.5, .5, .5);
-		rotationMatrix.rotate(Math.toRadians(arc.facing==2?180: arc.facing==4?-90: arc.facing==5?90: 0), 0,1,0);
+		int f = (arc.mirrored)?3:2;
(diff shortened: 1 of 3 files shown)


BluSunrize:
That's a weird way of being broken...why does it only affect 2 orientations?

BluSunrize:
Ah, I found the issue, it's actually in the scaling used for mirroring. If you don't mind, I'll fix this on my end

cobra:
It's the combination of rotation and scale(1,1,-1) for mirroring. Changing the scale parameters would also work, but that's not as simple as the one comparison for the rotation. Look at the excavator, you've done it there ;)

cobra:
Whatever, there's more than one way of fixing this. I found this to be the fastest and most readable form, but if you have a different solution, go ahead.

btw: do you have plans to make the refinery and assembler rotatable?

BluSunrize:
Assembler, no, because it's direction is very clear. Rotating would also require changing the pos value on every component of it.
Refinery I considered allowing mirroring, but the structure is symmetrical, so I never did.

BluSunrize:
fixed by https://github.com/BluSunrize/ImmersiveEngineering/commit/951ea821f125dedc889f2eae2b4f2c80e5a72e89

cobra:
Assembler: thought so, it'd be cosmetic only with the sides doing nothing.
Refinery: the redstone control panel could be mirrored to the other side, but I don't see a situation where this would be absolutely necessary.

Issue #493, opened 2015-10-09
Silos do not form / are broken(?)


theunkn0wn1:
Issue: Silos will not form or render in world.

When I build a silo or tank correctly, i cannot use the hammer to form them. I have rebuilt each structure exactly as they are shown in the book several times without success. Also: Existing tanks in my test world no longer render / function as storage devices

Built against commit f0a24dabdb84c81d0e114d47b55f3e3e4a68ba52
Screenshots below.

![image](https://cloud.githubusercontent.com/assets/3110986/10407774/8625d8cc-6ea4-11e5-9e01-8c97c34ba7a7.png)
![image](https://cloud.githubusercontent.com/assets/3110986/10407780/aa70ad56-6ea4-11e5-9318-19bfb0ba28cf.png)

danielkanjo:
update your optifine :) i had the same problem when i updated agricraft :) agricraft water tank was invisible :) i just updated forge and optifine and worked well :)

cobra:
I'm on 41411dc (one commit behind, base of my current branch, but that one commit is pipe code and my additions only affect the excavator) and everything silo-related does work absolutely normal. I remember having to click two (very rare) or three times (very, very, very rare) until the silo formed. Have you tried this without additional mods?

BluSunrize:
You need to click the center block on the 2nd level. aka the one you'd be looking at if you stand on the ground and face the thing centered.
It works fine on my current local and I haven't done anything with it in a bit.

cobra:
btw: that information is still missing in the manual.

BluSunrize:
Because it's so hard for people to experiment a little. Yeah, I know.
Also: IE 0.6 is still called "pre". I'm still writing the entries for the new stuff.

theunkn0wn1:
First off, no optifine.
Second, was totally unclear and kinda confusing where i had to click with the hammer to make it form. and yeah, i ended up spamming every block on the tank several times before I found the right one.

Third,
The tanks form now, but i still have two glitched silos in my world that won't break and won't render.

BluSunrize:
no matter if they won't render, the blocks are breakable. So I'd say, just remove the blocks. Closing this for now.

theunkn0wn1:
The silos in question do not function either. I can break individual blocks
but the MB persists. The storage functionality does not work and the
contents are lost. It's a really weird edge case caused by updating from
one 0.6.0_pre build to another... might bee worth investigation. Close it
[...]

PR #506, opened 2015-10-15
Comparator outputs for silos and tanks

malte0811:
Description is in the manual.
Also fixed silos not being createable, but that was caused by a commit pushed after #493 was reported :)


BluSunrize:
Umm. pulling this would revert all changes made to loss rates, remove the packet system, etc.
Please commit against latest.

Issue #514, opened 2015-10-17
Assembler issues


theunkn0wn1:
I have been playing around with the assemblers and have come across a few issues. 

1. Players can stack Assemblers
 * If several Assemblers are placed one after another without spaces inbetween, all available crafting opperations along that chain will execute simultaneously
 * Not sure if this is a feature or a glitch, @BluSunrize?

Example: I have a setup with three assemblers, each take the input of the previous to craft with:
![image](https://cloud.githubusercontent.com/assets/3110986/10560924/2fdd236e-74ce-11e5-9337-e0bb286fa542.png)
First and third assemblers
![image](https://cloud.githubusercontent.com/assets/3110986/10560930/487c3982-74ce-11e5-8ce8-3da33bc994d2.png)
second assembler
![image](https://cloud.githubusercontent.com/assets/3110986/10560934/5a785cb0-74ce-11e5-88b2-7d3a11b11b6e.png)
If I provide one enderium ingot to the first one, the enderium nuggets immediately output from the third assembler. I expected at least a noticeable delay as the output of the first Assembler must be transferred to the second, then that to the third. the transfer should be quicker than it currently is.

 [2] It appears the assemblers are munching on items.
  * With the same setup as above, i provided 63 Enderium nuggets to the first assembler and let it loop a few times. After several itterations, I lost enderium nuggets. To be exact, one itteration I lost one nugget. (total of 62 recovered) And no, its not on the ground or stuck in one of the machines, its just lost. After a second iteration two nuggets were lost. and the total nuggets lost appears to rise with every iteration :/
  * This also occurs with vanilla ingots\nuggets

IE 0.6.0_pre build 1f5d0aa

PS, can we please get an option to disable redundant crafting within one assembler (like I am doing above)?

BluSunrize:
1: The only reason it appears instant to you is because you managed to stack their ticks. The three internal recipes of an assembler are supposed to happen after eachother and then the output is moved to a valid inventory. Another Aseembler is a valid inventory and if it ticks immediately after the first one, it'll appear instant to you.

2: I'll need to look into that one, no idea what would cause that.

theunkn0wn1:
for my first issue, could you please change it to the items move over their output conveyor rather to the valid inventory directly outside the output? It would be really cool to see items flowing from one assembler over they conveyors to the next one without magically jumping a half of a meter before rendering.
[...]

PR #524, opened 2015-10-21
stopped conveyor belts voiding items, fixes part 2 of #514

malte0811:
Calling `setStackInSlot` on the client is not a good idea.
(It was pretty much impossible for you to see this commit before pushing, I pushed 4 minutes before you did :))


BluSunrize:
thank you =3

Issue #523, opened 2015-10-21
Pipes crash server when receiving fluid


cobra:
I don't know whether #513 is really fixed due to a lack of reproducibility, but at least the attempted fix for this does something¹ every time and very reproducibly. Also, it's much more effective by crashing the server instead of the client and making worlds unusable. How to enjoy this new feature:
- Load world with tanks connected to pipes: instant crash on server start. World cannot be opened.

or:
- Create new world
- build IE tank, fill with some liquid. Alternative: be lazy and use a TE portable tank set to output.
- connect IE pipe(s)
- apply redstone to bottom tank block
- crash, world cannot be opened again.

version: current master, 73552ec
````
java.lang.ClassCastException: blusunrize.immersiveengineering.common.blocks.metal.TileEntityFluidPipe$DirectionalFluidOutput cannot be cast to java.lang.Comparable
	at java.util.concurrent.ConcurrentSkipListMap.cpr(ConcurrentSkipListMap.java:655)
	at java.util.concurrent.ConcurrentSkipListMap.doPut(ConcurrentSkipListMap.java:835)
	at java.util.concurrent.ConcurrentSkipListMap.putIfAbsent(ConcurrentSkipListMap.java:1962)
	at java.util.concurrent.ConcurrentSkipListSet.add(ConcurrentSkipListSet.java:241)
	at blusunrize.immersiveengineering.common.blocks.metal.TileEntityFluidPipe.getConnectedFluidHandlers(TileEntityFluidPipe.java:71)
	at blusunrize.immersiveengineering.common.blocks.metal.TileEntityFluidPipe.fill(TileEntityFluidPipe.java:122)
	at blusunrize.immersiveengineering.common.blocks.metal.TileEntitySheetmetalTank.updateEntity(TileEntitySheetmetalTank.java:76)
	at net.minecraft.world.World.updateEntities(World.java:2160)
	at net.minecraft.world.WorldServer.updateEntities(WorldServer.java:515)
	at net.minecraft.server.MinecraftServer.updateTimeLightAndEntities(MinecraftServer.java:703)
	at net.minecraft.server.MinecraftServer.tick(MinecraftServer.java:614)
	at net.minecraft.server.integrated.IntegratedServer.tick(IntegratedServer.java:118)
	at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:485)
	at net.minecraft.server.MinecraftServer$2.run(MinecraftServer.java:752)
````

[...]

PR #534, opened 2015-10-26
Stop using ConcurrentSkipListSet

malte0811:
#523 and the gist i linked there contain additional information. Also fixes a stupid NPE.


BluSunrize:
wasn't ConcurrentSkipList thread-safe already? I can see why you changed it for the fluid pipes, but why is it getting changed for the ENet as well (sorry, this is my lack of proper java knowledge limiting me .~.)

malte0811:
Performance and convenience. `ConcurrentSkipListSet`'s are sorted, therefore write access has a runtime of `O(n*log n)` (or worse). You don't use the sorting feature anywhere, so I replaced those Sets with a different implementation that is thread-safe while not being sorted. This also means that you don't need to implement `Comparable` (and also can't forget to do so  :smiley:) when you don't need to sort something. If this solution had been used from the start, half of #412 and #433 would not have been necessary

BluSunrize:
dunno what I'd do without you guys. Thanks =3

PR #535, opened 2015-10-26
Create de_DE.lang

XFactHD:
German translation for everything, except the special revolver names, the skyhook, some subtitles in the manual and a command explanation where I couldn't find a good and understandable translation.


BluSunrize:
Awesome, thanks! :D

PR #556, opened 2015-11-02
fix fluid pipe directional output check

cobra:
Check the same direction we're outputting to.

This will fix setups like this one

Also put another critical fix in this one, fixing a fluid dupe bug when rightclicking IE tanks with buckets with placable fluids in them.
And a minor improvement: use sneak-rightclick with the hammer on the pump to change the mode of the opposite side. Very useful for pumping out of full-size and multiblock tanks.


BluSunrize:
ALL the fixes!

Thanks =3

PR #597, opened 2015-11-05
Work with all Forestry versions, using API. Add Forestry to gradle

mezz:
This should work with Forestry 3 and 4.

I added Forestry to gradle since it's convenient to have once it's there, and I knew exactly how to add it since I've done it before heh.
If you're not using gradle or something you can just keep `ForestryHelper.java`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,13 @@ buildscript {
     }
 }
 
+repositories {
+    maven {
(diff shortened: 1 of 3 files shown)


BluSunrize:
Yeah, I don't do Maven things, so I'll pull this manually. Didn't know the API allowed access to those variables^^

PR #613, opened 2015-11-07
New raytrace algorithm

malte0811:
since the vanilla one sometimes does not recognize a 4 meter thick floor.
Please don't pull this until @cobra confirms that he did not find any bugs.


malte0811:
Just added a new commit that prevents floodlights from being turned on and off to quickly (redstone clocks) because raytracing 8 32 block rays every other tick is bad for the server

PR #618, opened 2015-11-09
yet another bugfix collection

cobra:
this will fix some known and even more unreported bugs, mainly blocks not getting rerendered omn their own after changing them for other players on the same server. Example: placing another slab on top of an existing one to form a full block was only visible to the player doing this until placing another block.

Unrelated bugs found, but not fixed:
- items in the workbench aren't rendered for other players, block updates don't help.
- the sorter will always tell you "oreDict disabled" even when it is enabled. Exception: you're the player who just set it and you haven't unloaded it. Logging out and in again (or unloading the chunk, I guess) will reset the text overlay, but not the behaviour. I didn't change anything on the sorter.

Note: as any IImmersiveConnectables have their own packet handler, I haven't used the descPacket detection of the the other blocks here (wouldn't work, is always false) but the FML side detection by thread name.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/TileEntityIESlab.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/TileEntityIESlab.java
@@ -15,6 +15,8 @@ public class TileEntityIESlab extends TileEntityIEBase
 	public void readCustomNBT(NBTTagCompound nbt, boolean descPacket)
 	{
 		slabType = nbt.getInteger("slabType");
+		if(descPacket)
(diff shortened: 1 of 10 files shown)


BluSunrize:
Why side check? descPacket would have been much clearer and avoid unnecessary block updates, no?

cobra:
don't ask me why you wrote that TileEntityImmersiveConnectable onDataPacket method like that and called readFromNBT directly, which is calling readCustomNBT with descPacket=false :P I didn't want to start fiddling around there atm. As I said, this whole blockUpdate thing probably needs a more elegant, generic, and more automatic solution than this make-it-work fix.

Issue #551, opened 2015-11-01
Floodlight wire connection weirdness


theunkn0wn1:
Issue: It is possible to place LV wires through walls with a certain placement of floodlights. Pictures below show the setup:

![image](https://cloud.githubusercontent.com/assets/3110986/10867364/56955dc0-801b-11e5-8c85-968486e54c41.png)

![image](https://cloud.githubusercontent.com/assets/3110986/10867369/7b4e5b76-801b-11e5-8727-aef1eb206c97.png)
![image](https://cloud.githubusercontent.com/assets/3110986/10867372/8eb6fbe6-801b-11e5-9565-81efbfa844a2.png)

And a practical implementation...
![image](https://cloud.githubusercontent.com/assets/3110986/10867385/3502ae8c-801c-11e5-9a7a-44743c2d5243.png)

By starting at the lower flood light and connecting it to any of the upper ones, a valid connection is formed. This is unexpected, as there is no LoS between the connections due to the existence of a 2 meter thick solid wall between them.

IE build cf77da7
Bare-minimum MC instance.

cobra:
That's interesting. I'll investigate, I was the last one to touch the raytracing code afaik - but floodlights didn't exist then.

cobra:
I can't reproduce all of these cases, but your last one works in some orientations.
![](https://cloud.githubusercontent.com/assets/163331/10868214/0791a1ea-8087-11e5-850c-5b7a069e0981.png)
North is up, the lanterns with the red clay are the ones I can connect from the floodlight in the middle.

cobra:
The raytraceOffsets are a bit off, but even after finding these points I'm still confused how minecraft's algorithm doesn't collide with all the stuff which should definitely block. The raytrace is always returning null in these cases.
I once again wish I could just render the raytrace path. If anyone wants to join the bug hunt, feel free to do so.

BluSunrize:
I'll file this at low priority for now.

In fact, I'd rather get your PRs in @cobra. I was kinda hoping to maybe wrap up IE today or in the next few days.
Also, @cobra, you should really hop onto IRC some time (#BluSunrize on esper.net). It's just so much more convenient for discussions. =P

cobra:
I'm 99% sure that this bug can be found somewhere in the getRaytraceOffset methods for both the lantern and the floodlight. Which are looking a bit scary. (hey, Halloween is over!)

BluSunrize:
[...]

PR #629, opened 2015-11-10
fix connection offsets for floodlight and lantern

cobra:
second part of #551, respects the orientation of the floodlight base, makes wiring through the empty space of fences, posts and slabs possible, just like other connectors. Uses the same coordinates for raytracing and rendering now that the raytracer can ignore the starting blocks.

Also adds calls to super methods for writing/reading custom NBT data.

before:

after:

Note: the raytracer has a bug and hits blocks outside of the area it should check, so some connections might still fail until this is fixed.


cobra:
hijacked this PR to include a completely unrelated fix for wooden poles.

PR #643, opened 2015-11-13
fix creative capacitor texture

cobra:
there was one copper pixel per texture in these files, I changed it to purple (same colour as the other textures for the same place).


BluSunrize:
huh, I was sloppy.

PR #647, opened 2015-11-14
Updated en_US.lang

theunkn0wn1:
Changed `transferrate` -> `transfer rate` in several places as the prior is not a word far as I can tell.

Removed unnecessary comma on line 551 as it was creating an awkward statement between `non-fluid` and `blocks`.


BluSunrize:
I was pretty sure the comma was correct, since it was a list of attributes?

theunkn0wn1:
No, it is not. If read aloud, that comma creates sentence fragments that make little sense on their own; stating `non-fluid` and `blocks` as separate things entirely even though the table below shows the two as directly related. By removing the comma as i have done, `blocks` is modified by `non-fluid` rather than completely separate things to which `non-fluid` **does not exist within the context**. My modification to line 551 makes the sentence work and make sense to the reader. **`... other, non-fluid blocks,...`** vs **'...other, non-fluid, blocks...`**

 Also, I haven't seen an list of attributes in English that only has two entries that does not use "and/but/or/etc." to separate the two entries, not sure about other languages though.

Issue #227, opened 2015-07-17
Speed Loaded rendering is not refreshing until you open it on SMP.


Kubuxu:

BluSunrize:
I know, can't fix without custom packets. It's just how Minecraft works, I can't send item updates like that.

asiekierka:
Why not use custom packets? It's nowhere near as hard as one thinks, and lets you do a lot of things a lot more efficiently.

BluSunrize:
So far I've gone packetless in IE. It's a sort of challenge I set myself for this project. Go without packets for as long as I can ^^

PR #679, opened 2015-11-20
fixed desync caused by the speedloader

malte0811:
#227
This is possible now that @BluSunrize added a `SimpleNetworkWrapper`


malte0811:
I just added a second commit, the `IDeepStorage` integration was broken: it didn't write to the master but to the local TileEntity

BluSunrize:
\o/

PR #691, opened 2015-11-23
fix infinite loop on startup

cobra:
use equals for strings and compare against correct string to avoid
trying to insert over 6 million manual pages before running out of
memory and overloading GC.


BluSunrize:
That's what I get for adding that upper while loop at  0:30 am.

cobra:
That's exactly what I thought when I found this. :D

BluSunrize:
I thought it was a cool idea to handle update news based on lang file though ^^

mindforger:
i blame java to some point not having a default comparision operator for Strings

BluSunrize:
well I was comparing against the wrong thing too. xD

cobra:
Of course C++ operator overloading (and default arguments!) are really nice, but that still doesn't help you when you use the wrong value on the other side of it.

PR #715, opened 2015-11-28
Networking

malte0811:
Fixed a couple of desyncs and dedicated servers crashing on startup


malte0811:
wait, I'll add one unrelated commit

malte0811:
Merge if you want to

PR #727, opened 2015-12-02
fixed the redstone breaker sometimes not stopping to transfer energy

malte0811:
as discussed on IRC
also re-added one of my first changes to IE: made gradlew executable :)


BluSunrize:
I was hoping to not have it tick, but oh well...

PR #729, opened 2015-12-04
Remove redstone breaker signal relaying, minor changes

cobra:


BluSunrize:
This is for the better.

PR #765, opened 2015-12-11
fix hammer crushing

cobra:
return one corresponding dust instead of a silly amount of copper ingots


BluSunrize:
well THAT explains a lot.

PR #791, opened 2015-12-15
TileEntity fixes

malte0811:
went through all tile entities and fixed some bugs, improved performance in some cases. Also fixed a potential NPE in the silo, although the case in which it would have occured would imply that the silo was messed up before the NPE happened. Other stuff is in the commit notes


BluSunrize:
Well it's a lot less changes than that giant getTileEntity rewrite, so let's hope for the best! xD

PR #794, opened 2015-12-16
fixed wires sometimes not being dropped in singleplayer

malte0811:
This was caused by me in PR #763


cobra:
Errr… no. This will kill gregtech compat.

BluSunrize:
Yeah I like my GregTech compat =P

PR #842, opened 2016-01-04
Update ja-JP.lang

karakufire:
Add new feature entries, change translations.


karakufire:
I can't translate almost achievement titles and manual subtexts...
I haven't known their parody......what they come from?

BluSunrize:
Thank you =3

PR #852, opened 2016-01-07
Computer support

malte0811:
PR from @cobra and me. Please don't merge until @cobra confirms that this is ready to be merged.


src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityArcFurnace.java
```diff
@@ -700,4 +721,119 @@ public int getMaxEnergyStored(ForgeDirection from)
...
+
+	@Optional.Method(modid = "OpenComputers")
+	@Callback(doc = "function():int -- gets the amount of energy stored")
+	public Object[] getEnergyStored(Context context, Arguments args)
+	{
+		return new Object[]{master().energyStorage.getEnergyStored()};
+	}
+
+	@Optional.Method(modid = "OpenComputers")
+	@Callback(doc = "function(on:boolean) -- gets the amount of energy stored")
```
XFactHD: Shouldn't this be "switches the machine on or off" or something like that?

XFactHD:
The naming of the peripherals is not consistent and I also have a feeling that "IE:machinename" will be problematic as you don't put the component name into quotation marks when you call for a generic component and don't do the component.invoke() thing.

cobra:
Exactly: https://github.com/MightyPirates/OpenComputers/blob/master-MC1.7.10/src/main/java/li/cil/oc/api/network/SimpleComponent.java#L94-L96

malte0811:
Just pushed a commit that fixes the wrong documentation on `setEnabled` for the arc furnace and changes the OC component names to something that works. The main bit of that commit is that you now have to use an Adapter to make the machines work with OC. It is cleaner code-wise.

BluSunrize:
So is this good for merge now or still WIP?

malte0811:
I think that it is good to merge, but I don't know whether @cobra agrees with me.

Vexatos:
:+1: For using Drivers instead of Environments. Someone here knows how to use the OC API. Thanks.

malte0811:
That actually wasn't the reason for the change, making the network realize that the diesel generator had been mirrored was the problem. I sort of "hacked" around the "data conductivity" by
a) (this was @cobra's idea, which isn't hacky) only allowing a connection on a specific block on most multiblocks
b) making multiblocks that connect from everywhere (or at least multiple points) only connect from outside ([sample code](https://github.com/cobra/ImmersiveEngineering/blob/630a9cd41e6fb1b6b4cbb98a5bdc5254e958ee61/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityFermenter.java#L547-L553)). That wasn't an ideal solution, but it worked quite well.

BluSunrize:
crap I merged that into 0.7 xD

malte0811:
Just update your local 0.7 branch and run "git merge 0.7". Should work without a commit.

PR #866, opened 2016-01-12
Misc fixes

malte0811:


malte0811:
I just noticed that my fix for the fan animation contains a tiny bug: `preheater.lastRenderTick` has to be set to the current world tick regardless of `active` state. This just means that the fan is going to jump into a different position when you enable, then disable and then enable the preheater, so you can decide whether you want to fix that or not.

PR #869, opened 2016-01-13
Some bugfixes for the metal press

malte0811:
-sync on multiplayer was completely broken
-recipes weren't clearing properly, making it possible to make a gear with only 1 ingot
-fixed a small graphical glitch where the press would show the last output instead of the input for a short time


BluSunrize:
Thanks! =3

PR #872, opened 2016-01-14
Update zh-CN.lang ------ v0.7.0

Joccob:
When I test the lang file, I find that there are some of the links in the manual seem to be broken, for example, some of the links in "what's new in v0.6.0" and "Shaders" doesn't work quite well. There might be more ...I don't know if this is just me or not. But it happens for both English and Chinese and I hv no idea why this happens. Hope you guys can look into it and figure it out~

Best wishes


BluSunrize:
I fixed the link bugs with this:
https://github.com/BluSunrize/ImmersiveEngineering/commit/47b206161ba806fc7461b183b4dbcb5200cbfc0e

Issue #903, opened 2016-01-20
[minor] Search glitch


theunkn0wn1:
When a user uses the search function built into the Engineer's manual a minor error occurs. the page header for the search results inherits its value from the page originally searched from; this leads to inconsistency and potential confusion. Below are two identical searches initiated from two different starting pages, note how the title, even though its the same results, does not reflect the fact the page is of *search results* rather than whatever the old page was

![image](https://cloud.githubusercontent.com/assets/3110986/12440313/07405d20-beef-11e5-9a82-cf4b6c076d84.png)
![image](https://cloud.githubusercontent.com/assets/3110986/12440319/0fdfd190-beef-11e5-9840-5ea13d9d0ce2.png)

IE build 9af592c9166bbb6890b0008bcd905c22edc791c2

Issue #902, opened 2016-01-20
[minor] Multiple manual references with same name


theunkn0wn1:
Several entries when searched for, such as the Arc furnace and the, result in two identical results with no clear distinction which one is which without manually opening each result to find the exact page in question. I DO understand if browsing to the entries manually the distinction is clear however, for the sake of handy search feature, multiple entries with the same name is simply confusing to the player. 

![image](https://cloud.githubusercontent.com/assets/3110986/12440233/3e9dd334-beee-11e5-9226-126510d8d4d3.png)
![image](https://cloud.githubusercontent.com/assets/3110986/12440235/468b61f6-beee-11e5-8ae6-ba9a47348e21.png)

IE build 9af592c9166bbb6890b0008bcd905c22edc791c2

malte0811:
I didn't think of that when adding the computer documentation to the manual. I will possibly make another computer PR today (for capacitors), i could rename the entries then.

theunkn0wn1:
Perfectly understandable @malte0811.

Whilst we are here I have a question about the CC support you and cobra implemented, for IE's muiltblock structures where exactly do I need to place the computer upon the structure to gain access to the documented functions? Attaching a CC computer (golden) to any random side of an Arc Furnace leads to both unexpected and ugly results.

malte0811:
https://github.com/BluSunrize/ImmersiveEngineering/blob/master/src/main/resources/assets/immersiveengineering/lang/en_US.lang#L935
What do you mean by "both unexpected and ugly results"? It didn't provide any peripheral to me when i tried connecting somewhere else.

theunkn0wn1:
By unexpected i mean the peripheral calls exposed to the computer are not what I was expecting, either the correct calls or none at all where expected; instead only inventory calls are provided.

![image](https://cloud.githubusercontent.com/assets/3110986/12466691/26ce8062-bf8b-11e5-99b2-a7746894b808.png)
This is with proper CC placement directly on the RS port (thanks... was horribly unobservant when looking earlier...)
![image](https://cloud.githubusercontent.com/assets/3110986/12466707/4729c3a8-bf8b-11e5-879d-799222bebcc5.png)
improper placement using a well known (perhaps proper?) command to list peripheral calls ```m.listMethods()``` as it existed
![image](https://cloud.githubusercontent.com/assets/3110986/12466817/eb2c16e0-bf8b-11e5-94ed-85951a37299e.png)

[...]

PR #907, opened 2016-01-20
Capacitors for computers and fixes for #902 and #903

malte0811:
@BluSunrize didn't want computers to be able to set in-/outputs, so I didn't implement that.
The manual now just shows "Engineer's manual" as a category while searching.
Manual entries for computer support now have " (Computers)" behind their name to make them different from the normal entries.

I love that my variations of "I don't know any good subtitle for this" went into the release :smile:, the capacitor entry actually has a proper subtitle


BluSunrize:
Thanks for all that =3

PR #978, opened 2016-02-28
Shift the slot id's for OC

malte0811:
Also contains 2 small fixes for bugs I noticed while testing my changes.


BluSunrize:
Nice, thanks!

PR #982, opened 2016-03-01
Fixes for the drill

malte0811:
The drill no longer is able to mine obsidian ridiculously  fast by mining cobblestone next to the obsidian. I actually reported that bug on IRC in June :smile:
Fixed the drill animation in third person/multiplayer


BluSunrize:
Cheers

Issue #1030, opened 2016-03-23
Redstone breaker causes crash upon server join


ViciousVinnyD:
When a player joins the server and there is a redstone breaker nearby the client crashes. Getting another player to destroy the breaker before the client joins prevents this.

Actually placing a breaker does not cause this, only joining while there are existing breakers does.

Using ImmersiveEngineering-0.7.3

[08:32:04] [Client thread/INFO]: Connecting to 203.87.104.222, 25565
[08:32:23] [Client thread/INFO]: [CHAT] Welcome to the server!
[08:32:50] [Client thread/FATAL]: Unreported exception thrown!
java.lang.NullPointerException
	at blusunrize.immersiveengineering.common.blocks.metal.TileEntityRedstoneBreaker.getConnectionOffset(TileEntityRedstoneBreaker.java:48) ~[TileEntityRedstoneBreaker.class:?]
	at blusunrize.immersiveengineering.client.ClientUtils.drawConnection(ClientUtils.java:123) ~[ClientUtils.class:?]
	at blusunrize.immersiveengineering.client.ClientUtils.drawConnection(ClientUtils.java:117) ~[ClientUtils.class:?]
	at blusunrize.immersiveengineering.client.ClientUtils.renderAttachedConnections(ClientUtils.java:104) ~[ClientUtils.class:?]
	at blusunrize.immersiveengineering.client.render.BlockRenderMetalDevices.renderWorldBlock(BlockRenderMetalDevices.java:188) ~[BlockRenderMetalDevices.class:?]
	at cpw.mods.fml.client.registry.RenderingRegistry.renderWorldBlock(RenderingRegistry.java:118) ~[RenderingRegistry.class:?]
	at FMLRenderAccessLibrary.renderWorldBlock(FMLRenderAccessLibrary.java:53) ~[FMLRenderAccessLibrary.class:?]
	at net.minecraft.client.renderer.RenderBlocks.func_147805_b(RenderBlocks.java:296) ~[blm.class:?]
	at net.minecraft.client.renderer.WorldRenderer.func_147892_a(WorldRenderer.java:181) ~[blo.class:?]
	at fastcraft.a.o.a(F:86) ~[fastcraft-1.21.jar:?]
	at fastcraft.Hooks.t(F:147) ~[fastcraft-1.21.jar:?]
	at net.minecraft.client.renderer.EntityRenderer.func_78471_a(EntityRenderer.java:1187) ~[blt.class:?]
[...]

PR #1033, opened 2016-03-25
Misc fixes, once again

malte0811:
A lot of small fixes and one quite interesting change: Wires can transfer power through unloaded chunks again, this time without loading them (except when wires burn out, in that case the chunks are loaded for 1 tick, but that should not happen very often).
The fix for #1030 is a "probably" because my laptop can't handle the client alone too well, so I didn't even try to set up a server.


BluSunrize:
I don't know what I'd do without you! <3

Think we're good for a build tomorrow?

malte0811:
I don't have any more fixes or bugs that I would want to fix, so yes.

src/main/java/blusunrize/immersiveengineering/common/IESaveData.java
```diff
@@ -46,7 +50,13 @@ public void readFromNBT(NBTTagCompound nbt)
 				}
 			}
 		}
-
+		
+		Minecraft.getMinecraft().mouseHelper.ungrabMouseCursor();
```
malte0811: Don't merge this just yet, I left an `ungradMouseCursor()` call in that will crash servers.

BluSunrize: Roger that. I was gonan do it tomorrow probably?

malte0811: From what I know right know that should be ok. Just making the metal press a bit more fancy right now (Accept `<minimum input per mold>` in the first take instead of 1, would make it behave less weird in most situations, like in Dire's video today).

malte0811:
You meant to merge this a week ago :)

Joshwoo69:
yeah blu.. get to it :P

BluSunrize:
Just to point this out: Hammer has extensive rotation support in 1.8, so thanks for backporting a little of that, Malte =P

Also thank you for all your hard work, once again. Build coming up in a few moments.

Issue #1081, opened 2016-05-02
[1.8.9] Wire connectors doesn't transfer energy to Capacitors from EnderIO


Eufranio:
Nothing happen when connecting wire connector/relay in capacitor banks from EnderIO, then link them.
http://prntscr.com/azhyuf
http://prntscr.com/azhz0w

laz2727:
Try setting up the input-output sides.

AlterAkima:
We already did.

malte0811:
I am pretty sure that I know what is causing this, I will fix it tomorrow.

Eufranio:
Thanks :grinning:

BluSunrize:
Fixed in the latest release

PR #1090, opened 2016-05-04
Baked models for connections

malte0811:
Also a fix for #1081 and for raytrace/rendering offsets on the HV relay.
Don't poke the new rendering system, it will probably break :smile:


malte0811:
Don't merge this yet please, I just noticed that I used my "native" code formating in half this code and I forgot to add the electric lantern and the floodlight.

malte0811:
That's all I have for now. Should contain fixes for most reported bugs (except the two crashes that have crashlogs that don't make sense).

BluSunrize:
Sorry @malte0811 I created a merge confict because I got a report of that stackoverflow on discord and fixed it without checking if you did.
How do I resolve these conflicts again? .-.

PR #1114, opened 2016-05-09
fixed insulating glass being opaque and not letting through light

BluSunrize:
ported a few more bugfixes from 1.7.10


malte0811:
Interesting way of fixing this.

PR #1120, opened 2016-05-11
Mostly transformer fixes

malte0811:
Also changes source level to 1.8 and fixes weird IEOBJ models


BluSunrize:
Well yeah, but those hashes were built based on the deciding markers of the object. I don't need them to be == I just needed them to look alike. =P

PR #1139, opened 2016-05-17
Misc fixes

malte0811:
Details are in the commit notes.
One thing that isn't in the commit notes: Addon makers can make their blocks render wires by setting the model location to contain `models/block/smartmodel/conn_`. The part after that should be unique and has to be added as a key to `ConnLoader.baseModels` with the location of the real model as the value (This has to be done before models are loaded, e.g. in preInit). Right now only OBJ-models made with blender or software using the same standard for where `v=1` is will work, @BluSunrize can probably explain this in a less complicated way.


malte0811:
Only disadvantage of `git rebase` is that the commits are all dated to today. Otherwise a very nice thing.

Issue #1140, opened 2016-05-17
[1.8.9][bug] conveyor belt not moving items uphill or down properly


elkillo:
when using conveyor belts to move up and down hill the items tend to pop off when going downhill and get stuck hopping on the slop going up hill. is this a known issue or did i break something? i'm using the curse client and do not have fast craft installed. Version 0.8 - 26. To be clear the belts work perfectly fine when on the same plane but when going up or down the items almost seem to be falling through the belt when they reach the slope.

mindforger:
what other mods do you have installed? this is a pain that goes on since ever but nobody was able to find the culprit as with IE alone it always works fine ... must have to be some other mod doing stuff to floating items or something on the server side. 
Do you play MP or SP? If MP what kind of Server Base, clean forge or something like cauldron?)

elkillo:
i use clean forge from the curse client i play exclusively SSP with this pack. all mods are 1.8.9 and are current versions available on curse.

i use the following mods

better Foliage
blood magic
botania unoffical 1.8.9
buuildcraft
chameleon (lib mod)
Chisels and bits
clay world gen
Computercraft and computer craftedu
cooking for blockheads
endertanks
extra utilities 2
fast leaf decay
Immersive engineering
IC2
inventory tweaks
iron backpack
iron chests
journeymap
JEI
JEI resources
Mantle
Ore tweaker (to add extra iron there isnt enough for IE xD)
Pams harvestcraft
Psi
real time clock
Reliquary v1.3
SAO UI
shetiphianCore
simple harvest
storage drawers
thaumcraft
Tinker construct 2
waila

mindforger:
blood magic
XU
fast leaf decay
IC2
iron chests
Mantle
TinkersC
Pam's
Storage Drawers
waila
ThaumC
BC

thats all similaritys i could spopt with my pack so far

3TUSK:
@Amamiya-Nagisa can confirm it as well IIRC

elkillo:
i think once i get home ill disable all mods outside you list and try each one by one till i find the culprit. i can however tell you its not better foliage as i have tried disabling it already.

malte0811:
Happens in my dev-environment as well, if there are no blocks below the belt the items will fall through and if there is they get stuck on the edge of the first conveyor. I'll look into fixing this now.

elkillo:
i can confirm it does happen to players as well and malte's description is correct.

PR #1144, opened 2016-05-19
Conveyor belt fixes

### LINKED ISSUE #1142: [1.8.9][possible bug] dropping conveyors not dropping into some vanilla inventories below it
the Dropping conveyors don't seem to drop into some vanilla inventories, this includes vanilla furnaces top slot for input, trapped chests (which normal conveyors don;t insert into) to name a couple. They do however drop into hoppers are far as i can tell. Is this intended? or did a find a bug?

malte0811:
Fixes #1142 and #1140

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/api/tool/ExternalHeaterHandler.java
+++ b/src/main/java/blusunrize/immersiveengineering/api/tool/ExternalHeaterHandler.java
@@ -3,6 +3,8 @@ package blusunrize.immersiveengineering.api.tool;
 import java.util.HashMap;
 
 import net.minecraft.block.Block;
+import net.minecraft.block.BlockFurnace;
+import net.minecraft.init.Blocks;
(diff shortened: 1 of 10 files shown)


malte0811:
I pinged you in a comment on an experimental commit a few days ago, did you see that one?

BluSunrize:
Right yeah, I forgot to respond to it. I'd say go for it as long as it emulates vanilla placement logic entirely. Just so that it's not missing a call.

malte0811:
I think I copied the code from the vanilla logic, modified the broken parts and compressed some checks in the beginning. I don't know when you want to build, but I probably won't be at my computer until ~4pm, so if you want to build before that please check that I didn't delete anything. The relevant method is `ItemBlock.onItemUse`.

BluSunrize:
I'm gonna still do work on stuff as well, don't worry. Build won't be early.

PR #1156, opened 2016-05-22
There should be a funny joke about hearing here

xbony2:


BluSunrize:
These are things you could just message me about on IRC. They don't really validate pull requests.
Also, it's a config file, who really gives a damn?

xbony2:
I ain't got time for IRC \o/

PR #1161, opened 2016-05-24
Conveyor belt fixes

malte0811:
Details are in the commit notes.
Sorry for saying that your hopping system didn't work, it does, but I broke it with my first commit. The second one should fix it.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/api/tool/ToolboxHandler.java
+++ b/src/main/java/blusunrize/immersiveengineering/api/tool/ToolboxHandler.java
@@ -1,0 +1,68 @@
+package blusunrize.immersiveengineering.api.tool;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.BiPredicate;
+import java.util.function.Predicate;
+
+import blusunrize.immersiveengineering.api.energy.wires.IImmersiveConnectable;
+import blusunrize.immersiveengineering.api.energy.wires.IWireCoil;
+import net.minecraft.block.Block;
+import net.minecraft.block.state.IBlockState;
+import net.minecraft.item.ItemFood;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.ItemTool;
+import net.minecraft.world.World;
+
+public class ToolboxHandler {
+	private static final List<Predicate<ItemStack>> tools = new ArrayList<>();
+	private static final List<Predicate<ItemStack>> foods = new ArrayList<>();
+	private static final List<BiPredicate<ItemStack, World>> wiring = new ArrayList<>();
+	static {
+		tools.add((s)->(s.getItem() instanceof ITool?((ITool)s.getItem()).isTool(s):false));
+		tools.add((s)->(s.getItem() instanceof ItemTool));
+		foods.add((s)->(s.getItem() instanceof ItemFood));
+		wiring.add((s, w)->(s.getItem() instanceof IWireCoil));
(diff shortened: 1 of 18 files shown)


BluSunrize:
I refuse to pull this.
I clearly stated that the Tesla Coil isn't done AT ALL.
Don't fix stuff I haven't even finished yet.

malte0811:
Sorry, I'll remove the last commit later today.
You have contributed quite a bit, the whole 1.8 port!
How about this: I'll keep away from IE code for a week or so, I have a few other projects I want to work on (port the in-game nbt editor to 1.8, look into making an IE addon and one non-MC project).

BluSunrize:
I didn't want to come across like that. I honestly, truly appreciate your support. You're far smarter in terms of well-performing code than me.
I'll port some of the fixes in here (like the transportUp and Down stuff from the conveyors and other wonkyness) and I'll finish the Tesla coil. I actually meant to commit last night but forgot, so that's gonna be one giant commit xD

malte0811:
I removed the last commit and the double fixes from the commits before that, added something fancy instead! Github processing is being a bit slow right now, so it might not be visible yet.
Thanks for the compliment!
I assume you already know this, but tesla coils are being quite weird right now, shooting lightning from a meter away from the primary circuit. The lightning animation looks nice, though I would have preferred something with a bit of branching, but that is just my personal oppinion.
Btw, what happened to the licence thing you talked about on IRC some time ago?

BluSunrize:
Tesla Coils are still getting tinkered with yeah. The rotateability broke them again .~.
Oh btw. Getting damaged when falling onto a conveyor doesn't seem like a bad idea honestly. if people want jump cushions, they should use...well, jump cushions, don't you think? xD

malte0811:
The fall damage the commit is about is something that is most certainly unintended: as a player in survival mode, get onto a vertical conveyor belt and let it hop you up and down for some time (e.g. 30 seconds). Now walk out of the belt, onto the ground that is 1 block below and be surprised about having fallen from a high place.

BluSunrize:
OH! Fair point. I thought it was something about normal conveyors acting to break your fall xD

src/main/java/blusunrize/immersiveengineering/api/tool/ToolboxHandler.java
```diff
@@ -0,0 +1,68 @@
...
+import net.minecraft.world.World;
+
+public class ToolboxHandler {
+	private static final List<Predicate<ItemStack>> tools = new ArrayList<>();
+	private static final List<Predicate<ItemStack>> foods = new ArrayList<>();
+	private static final List<BiPredicate<ItemStack, World>> wiring = new ArrayList<>();
+	static {
+		tools.add((s)->(s.getItem() instanceof ITool?((ITool)s.getItem()).isTool(s):false));
+		tools.add((s)->(s.getItem() instanceof ItemTool));
+		foods.add((s)->(s.getItem() instanceof ItemFood));
```
BluSunrize: Those are lambdas right? I know little about them, but fancy. And from the looks of it you have TCon integration in as well.

Where you listening when I mentioned on stream that I need to have that for the toolbox? xD

malte0811: Both yes! I heard you say that on stream and thought that it would be a real nice use case for lambdas. It's the first time I have actually used them, but they are quite nice for this kind of thing.

PR #1176, opened 2016-06-04
Fix ie not correctly using the ic2 api

estebes:
This is a pr to fix an incorrect usage of the ic2 api which cause some ic2 machines to not accept power from ie connectors.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/util/compat/IC2Helper.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/util/compat/IC2Helper.java
@@ -1,5 +1,7 @@
 package blusunrize.immersiveengineering.common.util.compat;
 
+import ic2.api.energy.EnergyNet;
+import ic2.api.energy.IEnergyNet;
 import ic2.api.energy.event.EnergyTileLoadEvent;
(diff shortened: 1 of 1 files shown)


malte0811:
What machines? And using IC2 experimental or classic?

estebes:
ic2 experimental but ic2 classic uses the ic2 exp api so it will most likely work just fine. try it though.

Doken44:
IE 52 still has the same or very similar issue with Industrialcraft-2-2.6.105-ex110 (a release version)
I'm a nub. Has this been implemented yet?

ArukaAlter:
It doesn't do EU transfer since 1.8, and that's intended.

BluSunrize:
Correct. There is no EU support in 1.10, and it will stay that way.

chaos234:
@BluSunrize - I think that you have your reasons for not implementing EU support at/after MC 1.8 but is it possible to merge this PR for the 1.7.10 version?

Don't missunderstand me but merging it shouldn't hurt anybody or am I false? Will be glad if you would merge it (if not already done in version 0.7.7) and I can also test it to proof it.

Will be glad if you can "jump over your shadow" since it is for 1.7.10 (which is unsupported).

Issue #1169, opened 2016-05-31
Suggestion: Faraday suits


malte0811:
For anyone who does not know what a Faraday suit is: watch a bit of this [video](http://www.youtube.com/watch?v=bTkjgBRiCsE&t=9m0s)

We talked about this on IRC briefly yesterday and the problem is as follows (assuming that there were no misunderstandings):

I want Faraday suits, mainly to be able to do kind of an HV show. @BluSunrize does not want the coil to become useless as a defense mechanism which would happen if the Faraday suit protected you from the arcs.

I think I found a solution which will make both possible: Adding a "low power" mode to the tesla coil that consumes less power (x0.5?), but has a lower range (3 block radius?) and allows Faraday suits to work due to lower current. This mode could be activated by shift-rightclicking the TC with an engineers hammer while the TC is turned off or doesn't have enough power to run (IRL: because fiddling with the primary circuit of an active TC is quite deadly, in-game: to prevent intruders from switching defense coils to low-power mode). Using a Faraday suit at normal power mode would result in massive damage to the suit and fire damage in addition to the normal HV damage to the player (because of the molten parts of the suit).

For recipes, aluminum dust or something like that might work well. The suit should take very little damage from low power tesla arcs, but a lot of damage from any other source.

Another thing that might be neat in relation to tesla coils would be fluorescent/neon tubes that can be held or placed in the world as entities. Fluorescent tubes start glowing when they are near an active tesla coil, not enough to light up the environment, but enough to look cool.

If this suggestion is accepted, I would like to implement it myself (apart from the model/texture which would look shit if I made one).

XFactHD:
I really love the idea with the fluorescent tubes, I once saw a show were two people played Star Wars with a red and a blue fluorescent tube over a tesla coil.

malte0811:
I have seen similar things, for example someone wielding 2 fluorescent tubes while being hit by 2 meter arcs in a faraday suit. My idea were colourable tubes (normal mc colouring or even RGB) that could be turned using shift-mousewheel. In 1.9 it would be a nice thing to add dual-wielding support for.

AEnterprise:
maybe make it so that you can't change the coils mode when it's receiving power at all? as you are working with live wires

that would also mean nobody can mess with it from the outside

malte0811:
I have implemented the Faraday suit, now I need to add the fluorescent tubes.
[...]

PR #1181, opened 2016-06-05
Tesla stuff

malte0811:
Faraday suits and fluorescent tubes. See #1169.
Please merge #1167 before this PR, that will *probably* cause less problems with git (This PR is based on the first part of the other PR, that is why there are 5 commits instead of 2)


BluSunrize:
Well then! Time to tinker with this =P

BluSunrize:
I assume it's intentional that ITesla entities don't have lightning striking them? They function by pure proximity?

malte0811:
Yes, that is intended.

Issue #1202, opened 2016-06-19
Suggestion: dimension blacklist for pump.


delet3r:
I realize I probably represent a small minority, but it would be great if you could blacklist dimensions with the Pump, instead of just making it run like an enderthermic pump to reduce lag from pumping nether lava.  BC pump works this way, you can just make it not work in the nether.  

The 'convert to stone' works well for players who like pumping nether lava but it is worse for those of us who pump out lava lakes near bedrock in the overworld.  Instead of making more ores visible, it covers them in stone.  Again, I realize my opinion is probably not the most common one.

malte0811:
There used to be a feature in 1.7.10 that would have helped you event though it was completely different from a blacklist, but it apparently got lost while porting to 1.8: Pumps could be individually set to produce cobblestone or not by shift-rightclicking them with an engineer's hammer (this only worked when 'convert to stone' was enabled). @BluSunrize did that feature just get lost or did you intentionally remove it? If it got lost accidentally I'll re-add it

BluSunrize:
It got lost, that is my bad x3
I totally forgot about it too. If you wanna port it back in Malte, be my guest.

PR #1211, opened 2016-06-23
Misc fixes

### LINKED ISSUE #1193: [1.8.9]CPU-Spikes/Lagg-Spikes after installing IE
Hello,

After playing a while with IE stuff, my game got laggs periodically and I noticed a strange CPU behaviour, I'll attach a Pic of the CPU-Load (http://fs5.directupload.net/images/160613/uniivzr3.png)

I'm not 100% sure if it's IE but it's quite likely cause it happens only after playing with IE items/blocks and it haven't happened before installing IE.

Removing or installing Optifine don't make any difference.

Here are the other mods I installed: http://prntscr.com/bfvkty

("thaumcraftServerHelper" is something from myself, It's not causing it, I already checked it!)

It happens on a Server as well as in SP

cad435

### LINKED ISSUE #1207: Light comes through concrete.
I litterly just went to 3 diffrent mod authors about this and was totally wrong. I hope this is correct...

I had a lot of lag in my world after I updated IE. At first I thought for some reason Pneumaticrafts etching acid was effecting my performance. Well, no.
My base is over an ocean. The entire floor is concrete. When I updated, light could go through concrete, so that is what was making me lag, was the updates under my base.

Unfortunately this means when you fix this, there will be more lag, but at least it will be fixed... >.<
:D

malte0811:
_(I need to think of better names for my PR's)_
Fixes #1207. May fix #1193. Makes #1202 less important. Fixes broken floodlight rotation rendering. Makes the pump suck up fluid blocks that are connected to the pump via flowing blocks

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/client/models/IESmartObjModel.java
+++ b/src/main/java/blusunrize/immersiveengineering/client/models/IESmartObjModel.java
@@ -33,7 +33,6 @@ import net.minecraft.client.renderer.texture.TextureAtlasSprite;
 import net.minecraft.client.renderer.vertex.VertexFormat;
 import net.minecraft.client.resources.model.IBakedModel;
 import net.minecraft.item.ItemStack;
-import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.util.ResourceLocation;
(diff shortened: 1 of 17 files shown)


malte0811:
The last commit breaks things, I'll push a fix once it is ready.

PR #1225, opened 2016-07-03
Re-add computer compat

malte0811:
Mainly just copied from 1.7.10 (For reference:  #852). I will check the code for errors tomorrow, just putting this here so anyone else who wants to look at the code can do so.
Also fixes filled buckets stacking in the refinery output slot.


malte0811:
I just fixed a few minor bugs in my previous commits and added computer support for the Tesla coil. I think this PR can be merged now.

PR #1231, opened 2016-07-11
Changelog in the manual and misc fixes

malte0811:
Most important fix is a memory leak in the model caches if a block state has missing (`Optional.absent()`) values.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/client/models/smart/ConnModelReal.java
+++ b/src/main/java/blusunrize/immersiveengineering/client/models/smart/ConnModelReal.java
@@ -183,16 +183,22 @@ public class ConnModelReal implements IFlexibleBakedModel, ISmartBlockModel
 			if (!(obj instanceof ExtBlockstateAdapter))
 				return false;
 			ExtBlockstateAdapter o = (ExtBlockstateAdapter) obj;
-			for (IProperty i : state.getPropertyNames())
+			for (IProperty<?> i : state.getPropertyNames())
 			{
+				if (!o.state.getProperties().containsKey(i))
+					return false;
 				Object valOther = o.state.getValue(i);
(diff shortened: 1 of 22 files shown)


src/main/java/blusunrize/immersiveengineering/client/models/smart/ConnModelReal.java
```diff
@@ -183,16 +183,22 @@ public boolean equals(Object obj)
...
 				return false;
 			ExtBlockstateAdapter o = (ExtBlockstateAdapter) obj;
-			for (IProperty i : state.getPropertyNames())
+			for (IProperty<?> i : state.getPropertyNames())
 			{
+				if (!o.state.getProperties().containsKey(i))
+					return false;
 				Object valOther = o.state.getValue(i);
-				if (valOther == null || !valOther.equals(state.getValue(i)))
+				Object valMine = o.state.getValue(i);
```
BluSunrize: shouldn't this just be "state." instead of "o.state"? Otherwise you're comparing the other to itself.

malte0811: You're right, didn't see that.
I won't be anywhere near my computer for the next 1-2 weeks, so I can't fix this right now. If you want to merge this before I am back, please replace the `o.state.` by `state.` for `valMine` and replace the parameter to `equals` in the next line by `valMine`. Since I can't test the fix, I recommend that you check the cache size after the changes and check that rendering doesn't go crazy.

BluSunrize: Yeah I am aware of this issue since i have addressed it for 1.9 and 1.10. I'll see too it, I hope to get the last 1.8 build out this weekend and go full up 1.9/1.10 after that.

PR #1267, opened 2016-08-11
Add villager name

aruneko:
I noticed villager's name is too long.
So, I named some villagers.


BluSunrize:
No. I don'T accept PRs on a branch that has no public builds. 
It's still WIP, of course stuff like that is a thing.

PR #1273, opened 2016-08-16
Hammer Default Durability from 100 to 250

MalkContent:
to be in line with vanilla iron tools
also 100 is really a bit low


BluSunrize:
Lolno? Random PR without even approaching me first to then change my default balance values? Noooope.

MalkContent:
please don't take this as me being impolite, I just thought "there's nothing to discuss. either he takes it or not" ^^

MalkContent:
Boy do i regret this. I was convinced that using the hammer to break IE blocks was lowering the durability.
My sincere apologies.

Issue #1281, opened 2016-08-18
[1.10] Wires do not render under certain conditions


mezz:
![2016-08-18_16 23 20](https://cloud.githubusercontent.com/assets/916092/17793956/4aaa725e-6560-11e6-8406-1364f835bf1d.png)
![2016-08-18_16 23 22](https://cloud.githubusercontent.com/assets/916092/17793957/4aaad488-6560-11e6-849b-cd0a0a4cbff6.png)

In the screenshots I take one step and the wire stops rendering. If I turn around so that the other connector is in view, the wire will start rendering again.
My best guess is that having my back to the connector culls it and the wire stops rendering as well. I can't seem to make the wire disappear when both connectors are in view.
This happens for all cables, just walk between two connectors with a wire overhead. For one of the directions you walk, the wire will stop rendering about halfway between the connectors.

mindforger:
attach a log please

mezz:
I was able to reproduce with just IE.

![2016-08-18_23 43 02](https://cloud.githubusercontent.com/assets/916092/17801231/a2aaf39c-659d-11e6-8fe9-353b8f5b27de.png)
![2016-08-18_23 43 04](https://cloud.githubusercontent.com/assets/916092/17801232/a2ab86b8-659d-11e6-8070-cf70de57d3c1.png)

Forced Crashlog (same one as the other issue):
https://gist.github.com/mezz/508381d8d1d101edc39dc5f6786d20d1

BluSunrize:
@mindforger tip: Mezz is Forgecraft crew. When he reports an error, he's good with it. =P
Also, yeah, known issue, seen it plenty on tlove's stream

mindforger:
yeah i know but if he doesn't attach logs, why should other do it then ... all a thing of equal treatment and beeing an exampole IMHO .. sorry for the bold undertone :P

ShadowDrakken:
Just a thought here, but is this even a fixable issue?

I think this comes from some of vanilla's optimizations that happened in 1.8 and again in 1.9 where it's culling render of blocks and entities that are outside the player's view. The same behavior has been showing up for EnderIO's travel anchors, and I suspect it's the render culling there too; I use Chunk Animator so when I'm inside and can't find the travel anchor I want, I step outside and watch the travel anchor as it renders and suddenly it works correctly.

Great for speed, no so great for rendering things that cross into the culled area...

KnightMiner:
[...]

PR #1290, opened 2016-08-22
Fix for #1281

malte0811:
Well, not really a fix. But something to reduce the impact. Each connector now renders half of each connection and adds one fading segment to the end of its half of the wire, as seen in the screenshot below. Screenshot


KnightMiner:
Can you post a picture of how it looks when both halts are together? Basically so we can see if there is any noticable difference.

malte0811:
There is a small difference, but I wouldn't notice it unless I was looking for it. You have to be quite close to the wire to see it and hemp rope is pretty thick IIRC.

BluSunrize:
although the fadeout is weird, it actually looks quite beautiful. I personally wouldn't mind a 1.10 PR on this =P
Oh btw., malte, for future reference, I bake the quads for the Wires with DefaultVertexFormat.ITEM now. Not block. I think it fixes the Black renders.

PR #1341, opened 2016-08-28
#1290 for 1.10

### LINKED ISSUE #1280: [1.10] Wires and Connectors render as black (lighting issue)
This is on FC with the latest 1.10 build.
Any wire or connector I place is black as in the screenshot.
Relogging does not resolve the issue.

malte0811:
- wires split rendering between both connectors. The split happens at a chunk border. (The transparent bit at the end doesn't look as good as it does in 1.8, I blame the changes to the vanilla rendering code from 1.8 to 1.10 for that)
- fixed wire lighting, closes #1280 (this needs to be done manually since I forgot to add it to the commit note)
- fixed inverted textures on half of a wire's faces (the black and gray bits did not line up properly)


BluSunrize:
That's awesome Malte, thanks! <3

PR #1408, opened 2016-09-08
Update zh_CN.lang

PeakXing:
Add new fields
Modify some fields


3TUSK:
Evidence of machine translation found.
* `table` -> 表格 (lit. form, sheet, table) is totally wrong in term of Lua. Lua's `table` is associative array; for Simp. Chinese, it is `关联数组`, or `链表` if my memory is still good. (`链表` often makes me think about `LinkedList` in Java though.) Personally I strongly *object* to translate `table` if referring to Lua's `table`. @Vexatos 
* `流体的未定位名字` Totally wrong. Unlocalized does not mean "not located". Also consider `unlocalized` vs. `delocalized`. 
* `当采样完成时返回真, 不然返回假` `true` -> 真, `false` -> 假, technically it **is** correct terminology in Chinese when true/false is referring boolean value; however, personally speaking, it looks weird. 

Calling some people @GWYOG @Mrkwtkr @Meow-J as I don't have enough time right now to clone the repo and check the full diff locally.

*Please, do most players' favor, do not use machine translator!*

Also @BluSunrize please don't merge this pull request - as I said, I am 100% confident about the evidence of machine translator. Experienced minecraft mod translator won't make such mistake.

CannonFotter:
Maybe you push it to a wrong branch....

P.S.(in Chinese,never mind)
不管您能不能看到这条，视而不见还是如何，我希望您能停止这种不负责任的行为，这是对mcmod中文本地化的一种伤害，翻译是被欢迎的，但不是这种翻译。(以及语言文件除非作者允许，自己加上作者信息并不好)

PR #1473, opened 2016-09-17
Breaking News: Rusty Coder Attempts Arc Furnace Input Fix

dualinfinities:
I set out to fix the way the input ports (in 1.7.10) put additives like copper dust or obsidian into the smelting slots by separating the main slots from the additive ones. this may need cleanup and I may have used the wrong pos for the top inputs. tell me if anything is messed up, and I'll fix it. the texture for the top could be improved by replacing the blue port rings with green and purple ones to distinguish them, but the positioning is a "good enough" indicator.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityArcFurnace.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityArcFurnace.java
@@ -641,11 +641,11 @@ public class TileEntityArcFurnace extends TileEntityMultiblockPart implements IE
 	@Override
 	public int[] getAccessibleSlotsFromSide(int side)
 	{
+		final TileEntityArcFurnace master = master();
 		if(!formed)
 			return new int[0];
-		if((pos==86||pos==88) && side==1)//Input hatches on top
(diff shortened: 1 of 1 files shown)


chaos234:
@dualinfinities 
I am not shure why you did this because I have no problems with it.

The pos-values are correct and yes the import graphics can be changed to green and purple if this is possible.

But at the end splitting this should not make any difference how it currently works.

dualinfinities:
it's because BOTH inputs were treated as THE SAME INPUT in the 1.7.10 branch prior to my fix. honestly it doesn't matter anymore since 1.7.10 is pretty much dead.

dualinfinities:
didn't matter which hole you put it in, it'd be sent wherever. that's a huge issue with automation, b/c most reagents are also smeltables, and would frequently get sent to the main slots if they arrived before the base did.

kane-thornwyrd:
Blu stated that the 1.7.10 __IS__ dead and won't receive anymore fixes afaik.

dualinfinities:
yeah, about what I expected. I'm not playing MC at the moment, so someone should check if the current builds have the same issue.

BluSunrize:
1.10 has had this fixed for ages and 1.7 is not getting an update, so I might as well close this :V

Issue #1453, opened 2016-09-14
[0.10-40] FPS drop when looking at waterwheel


guachitonico:
I run minecraft 1.10 at about 130FPS overall. However, when I look at my waterwheel setup (3 waterwheels mostly surrounded by wooden planks, connected with IE wiring), FPS drop to about 40 (the drop is about 40% fps)
The F3 debugger shows that the 'game renderer -> Entities -> Block entities' is the responsible of the slow down.

I will be happy to do more debugging, if needed.

malte0811:
What version of IE are you using? If you are using a version before build 39 (0.10-39), try updating. That version might fix or at least reduce the lag they cause.

guachitonico:
I'm using 0.10-40

miuirussia:
Confirm that issue in latest build from source code

BluSunrize:
Yeah. I can't help it. New rendering in TESRs isn't pretty.

RobertLSnead:
I have the same problem! :(  I was not sure what was causing it till I checked several posts!  Thanks for sharing!  Hopefully we will see a fix for this?

BluSunrize:
You probably won't. I can't really fix this any further than it is.

guachitonico:
Surely it's just one or two functions taking up most of the time, right? During the next days, if not tomorrow, I'll try to pinpoint the bottleneck.

BluSunrize:
Yeah, It's the function that renders the model. I can easily improve FPS by having it not render.

guachitonico:
It would be an easy fix for sure :smiley: 
Jokes aside, I can't recall this happening in previous IE versions, that's why I find it weird.

BluSunrize:
Because "previous versions" were 1.7 and rendering changed entirely in 1.8.

AEnterprise:
could the forge animation api help in making it less intensive or are you already using that?

malte0811:
I have done some optimizing on the waterwheel render (still need to port those changes to the windmills), see the commit above. It isn't perfect, but with 20 waterwheels in an otherwise empty superflat world I get ~15 FPS on master and ~30 FPS on 05743f120016858ae75a02fa25a7b6bc3f45571e.

BluSunrize:
It's not the animation that does it. It's the rendering of the quads.

guachitonico:
Great to see at least some optimization can be done for it.

enzyme-a113-2:
[...]

PR #1509, opened 2016-09-25
Fix for water wheel FPS

malte0811:
See #1453
Also adds some parameters to API methods that I need for Industrial Wires, fixes the `<ERROR>` message in WAILA if WAWLA is installed and makes windmills and waterwheels a lot less stutter-y.
The first 2 commits can basically be ignored.


malte0811:
@BluSunrize this should be ready to be merged now. The normal windmill is currently rotating weirdly, but that is because the mill in the OBJ file is facing in a different direction. It would be very nice if you could do the rotating because I don't really know blender...

PR #1518, opened 2016-09-27
Removed debug code

xbony2:


BluSunrize:
Already changed locally. Also, it's really pointless to do PRs for small shit like that.

xbony2:
You probably got everything, but there's a [fair amount](https://github.com/BluSunrize/ImmersiveEngineering/search?utf8=✓&q=system.out.println&type=Code) of other debug code, so make sure you didn't miss anything ^_^

BluSunrize:
well then maybe PR all of it out instead of doing a single line PR? ;D

xbony2:
I was going to do that, actually, but then I'd have to pull up eclipse and everything and that would be too much effort :P

elifoster:
If you use IntelliJ IDEA you can configure it to warn you for any System calls (or maybe it's System.out references).

Issue #1520, opened 2016-09-30
Immersive Engineering Server Crash on version .42 "ArcRecyclingThreadHandler"


blkwind:
I keep getting this crash on our server and cannot figure out the cause of the issue. If I try booting a fresh world it crashes right away. I have been able to get it to boot without a crash by adding it to the server after the world has been created however if I reboot the server the crashes start again.

"Description: Watching Server

java.lang.Error
	at blusunrize.immersiveengineering.common.crafting.ArcRecyclingThreadHandler.doRecipeProfiling(ArcRecyclingThreadHandler.java:103)
	at blusunrize.immersiveengineering.ImmersiveEngineering.serverStarted(ImmersiveEngineering.java:186)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at net.minecraftforge.fml.common.FMLModContainer.handleModStateEvent(FMLModContainer.java:597)
	at sun.reflect.GeneratedMethodAccessor8.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at com.google.common.eventbus.EventSubscriber.handleEvent(EventSubscriber.java:74)
	at com.google.common.eventbus.SynchronizedEventSubscriber.handleEvent(SynchronizedEventSubscriber.java:47)
	at com.google.common.eventbus.EventBus.dispatch(EventBus.java:322)
	at com.google.common.eventbus.EventBus.dispatchQueuedEvents(EventBus.java:304)
	at com.google.common.eventbus.EventBus.post(EventBus.java:275)
	at net.minecraftforge.fml.common.LoadController.sendEventToModContainer(LoadController.java:239)
	at net.minecraftforge.fml.common.LoadController.propogateStateMessage(LoadController.java:217)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
[...]

PR #1524, opened 2016-09-30
Fix for the non-descript error on startup

malte0811:
#1520
Who thought that making the server shut down in such a way was a good idea?!

I also turned on caching for conveyor models, they seem to be working fine. If you turned them off for a reason, please tell me, I'll turn it back off.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityExcavator.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityExcavator.java
@@ -475,4 +475,12 @@ public class TileEntityExcavator extends TileEntityMultiblockMetal<TileEntityExc
 	{
 		return null;
 	}
+	@Override
+	public void disassemble() {
+		super.disassemble();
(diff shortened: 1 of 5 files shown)


src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityExcavator.java
```diff
@@ -475,4 +475,12 @@ protected IMultiblockRecipe readRecipeFromNBT(NBTTagCompound tag)
...
 	}
+	@Override
+	public void disassemble() {
+		super.disassemble();
+		BlockPos wheelPos = this.getBlockPosForPos(31);
+		TileEntity center = worldObj.getTileEntity(wheelPos);
+		if (center instanceof TileEntityBucketWheel)
+			worldObj.addBlockEvent(center.getPos(), center.getBlockType(), 0, 0);
+	}
 }
```
BluSunrize: Funny, I had the same idea earlier when I was doing the permutations. Cheers =P

PR #1525, opened 2016-09-30
Update zh_CN.lang

3TUSK:
This pull request redoes #1336 #1408 and #1409. 
Also, [en_US.lang line 1282](https://github.com/BluSunrize/ImmersiveEngineering/blob/master/src/main/resources/assets/immersiveengineering/lang/en_US.lang#L1282):
> §lname§r: the internal name of the item, i.e. "minecraft:cobblestone" for cobblestone<br>§lsize§r: the size of the item stack<br>§lnameUnlocalized§r: the unlocalized name of the item, i.e. "tile.stonebrick" for cobblestone.

Personally speaking, it is better to use "e.g." here rather than "i.e.". "i.e." does not mean "for example", but "that is" or "that said" ("id est" in latin).


BluSunrize:
Yeah, I prefer e.g. as well, mostly because I'm tired of people pinging me with "ie" on IRC. x3
No but seriously, e.g. sounds better, personally.
@malte0811 you wrote the computer documentation. Mind if I change that?

malte0811:
Yes, e.g. sounds better. I don't know why I put i.e. in the first place.

3TUSK:
Also I need to ping @qianzha for translation of entries of Blood Magic compatibility... for now I commented that entry out.

PR #1603, opened 2016-10-16
Breaker switch rotation

malte0811:
I wanted to make them rotatable, but that was a mess with the models (might be just as bad now, but I don't want to rework everything again). Please look through this code and correct any derps I made, I am quite tired.


BluSunrize:
I'll address it tomorrow, build and released until after this is fully implemented.
Thanks for the PR ^^

PR #1673, opened 2016-11-11
Get CraftTweaker API from maven using gradle

mezz:
To get the project working I had to grab CraftTweaker, this PR will do it automatically using gradle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ out
 *.iws
 *.iml
 .idea
+classes
 
 # gradle
 build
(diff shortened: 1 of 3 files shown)


BluSunrize:
@mezz at the time of me (re-)implementing crafttweaker compat, there was no maven for it.
And yeah, clarification about code changes please.

mezz:
Yeah the maven is new, and also hard to find. 

The code changes are because whatever helper function you were using before is not in the minetweaker API, it's part of another library. Instead of depending on that library for just two functions, I asked Jared and he said you can get the contents of 'IItemStack' and 'IFluidStack' this way.

Issue #1634, opened 2016-10-24
ToDo: Add an offset/partial texture system to Item Shaders


mezz:
All these are big blank textures for me:
immersiveengineering:revolvers/shaders/revolver_sponsor
immersiveengineering:items/shaders/chemthrower_sponsor
immersiveengineering:items/shaders/drill_diesel_sponsor
immersiveengineering:items/shaders/railgun_sponsor

You can investigate for yourself using the newly released TextureDump mod:
https://github.com/mezz/TextureDump

<3

BluSunrize:
come on mezz, at least show me the sheet <3

BluSunrize:
Keeping this open as a reminder btw.
Fry's idea of using offset specifications isn't bad, but not as easily applicable as it sounds. Will have to mull this over and come up with a good design.

Refurbishing this into a ToDo: Add offset and partial textures to Shader system.

BluSunrize:
Shader Commit has happened, see
https://github.com/BluSunrize/ImmersiveEngineering/commit/da596e593c3ea28ae35c50ff2dad6baf1fd10471
https://github.com/BluSunrize/ImmersiveEngineering/commit/e449629ffd52329b55ee681a1c07f29c68aabb56
https://github.com/BluSunrize/ImmersiveEngineering/commit/267cf97eb74843702490162401dd4f84c0787353
https://github.com/BluSunrize/ImmersiveEngineering/commit/a5d513fefee97aeaa102461e1bdfbcf46eae77ae
https://github.com/BluSunrize/ImmersiveEngineering/commit/6249cb9077ff4a8cd3904b594cf36762091dc8c4

Issue resolved.

mezz:
Fancy! Looks really cool Blu.

PR #1676, opened 2016-11-12
Colorable balloons and misc

### LINKED ISSUE #1650: NPE when placing a normal conveyer belt in front of a vertical one
I assume they form some type of combined block to connect them. So I had a vertical one going up and the lowest block empty. When I placed a normal belt there it crashed and now my world does not want to load anymore (obviously if it's in the render method). 

I don't get he chisel detail there as there was nothing chiseled in this area (but maybe chisel just extends some components). 

Here is the log: https://gist.github.com/mrtopf/0622f9b2dee0db282c72db3c194194f7

malte0811:
I just realized I probably should re-add balloon coloring to the manual, but feel free to do that yourself if you want to merge now (I don't want to do a 1-2 line commit).
Since you are currently rewriting shaders (I guess because of #1634) I might take a look at re-adding balloon shaders once you push your changes.
I know you wanted to fix #1650 yourself, but that issue was preventing me from logging into my dev server, so I fixed it.


BluSunrize:
Yeah this commit is probably gonna break quite a bit.
Shaders got a massive rehaul. Might if we postpone it till I merge my shader rework into master?

malte0811:
I won't touch shaders until your new shader code is public (if that's what you are saying). The PR as it is does not do anything with shaders, just coloring. 
Btw, does the rehaul affect gameplay or just the code?

BluSunrize:
It changed the rendercode on smart models a little bit. That's why I don't wanna pull this yet. I think it'd be easier for me to fix merge conflicts on pulling your commit into my new render stuff rather than fixing the conflict of my new render stuff into this. =P

malte0811:
I'll merge master into this and check that there are no problems.
Would you be opposed to adding an API interface `RotationConsumer` or something like that to allow for custom blocks using rotational energy like the dynamo? It's just an idea for basic (low-medium power, low efficiency) EU/RF conversion in IndustrialWires.

BluSunrize:
I would not be opposed, but I don'T have time to write it atm. I can barely keep on top as is. If you wanna PR it, I'm all for it.

BluSunrize:
Otherwise though, this commit looks good to me. I'll go merge this, the Rotation stuff can happen later I think.
But I want to get a few of those fixes out.

BluSunrize:
In fact, I'll merge this in 2 hours, just in case there is any emergency things you wanna put in before.

malte0811:
I have some unpushed stuff, this includes the rotational energy interface. I'll push that stuff once I am at home, that should be before 4 pm.

malte0811:
Turns out the rotational energy was the only unpushed change I have that wouldn't require a lot of checking (I made some changes to wire models some time ago, but those would need to be tested before putting it into a release);
You can merge this now if you want to.

Issue #1692, opened 2016-11-19
Tesla Coils(?) crash the games sound manager


MalkContent:
I assume it's the tesla coils because i got a 5x5 square surrounded by them and the crash doesn't seem to happen without the tesla coils on.

IE 48
mc 10.0.2
forge x2122

crashlog:
http://pastebin.com/a2W9VWmj

edit: pastebinned the crashlog

AEnterprise:
github doesn't have spoilers but you can put the crash in a gist/pastebin and then put the link to it instead of the full log in your post

BluSunrize:
Next time, please read the contribution guidelines and spoiler your crashlog =)

I'm a little stumped on what causes this one though. Shouldn't be able to throw a concurrent modification exception here. Sounds for the tesla coil are only processed in its clientside update method

mindforger:
didn't you deal with that error in the past with the generator and crusher? i could not find the issues anymore but the report looked very similar

malte0811:
> Sounds for the tesla coil are only processed in its clientside update method

No they aren't. [This](https://github.com/BluSunrize/ImmersiveEngineering/blob/master/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityTeslaCoil.java#L280) call is in the `receiveMessageFromServer` method, which is probably called from the networking thread. I'll implement a thread safe way for that, shouldn't take too long.

BluSunrize:
fair enough.
Think you can get it PRed today?

malte0811:
If I don't find any other nasty bugs I should be able to make the PR today.

Issue #1683, opened 2016-11-17
Voltmeter crashing server


TheLegitimantis:
Sometimes, using a voltmeter on a machine with a large volume of RF moving through it will crash my server. Sometimes it happens and sometimes it doesn't.

Here is my server's crash report: http://pastebin.com/c2bQsaU6

This happens with capacitors and the excavator, mostly.

This happens with 0.10-48, Forge version 12.18.2.2114

malte0811:
Does this happen only when actually right-clicking with the voltmeter or at a random time when looking at a machine? (I would "prefer" the latter since I would know what is causing this in that case)

TheLegitimantis:
This happens when looking at the machine with the meter held in hand. I've never had an issue with right clicking with it.

PR #1694, opened 2016-11-19
Fixes for 2 multithreading bugs

malte0811:
One between client and networking (#1692), one between server and networking (#1683).


AEnterprise:
there is an easier and cleaner way, the server world and the minecraft client instanstance are both IThreadListener so you can add a scheduled task to them to be executed from the main thread

source: http://mcforge.readthedocs.io/en/latest/networking/simpleimpl/

PR #1752, opened 2016-11-29
[Potions] Fixes ArrayIndexOutOfBound crash

### LINKED ISSUE #1071: java.lang.ArrayIndexOutOfBoundsException: 36 	at net.minecraft.potion.Potion.<init>(Potion.java:68)
[crash-2016-04-30_02.10.34-client.txt](https://github.com/BluSunrize/ImmersiveEngineering/files/243494/crash-2016-04-30_02.10.34-client.txt)

I hope this helps. 
I can't get past the phase 3 of the starting process.

gnif:
The existing code extends the potionsType array by a fixed length (4), this is incorrect since minecraft uses the ID as the index in this table, if the ID is configured to some value larger then the array it will throw an ArrayIndexOutOfBounds exception.

This patch enlarges the array to the size required for minecraft to store the potion ID.

Fixes #1071

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/util/IEPotions.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/util/IEPotions.java
@@ -21,31 +21,41 @@ public class IEPotions
 
 	public static void init()
 	{
-		int l = 4;
-		extendPotionArray(l);
 		long potionUUIDBase = 109406000905L;
 
-		int potionID = Config.getPotionID(24, "Flammable");
+		int potionID = getPotionID(24, "Flammable");
 		flammable = new IEPotion(potionID, true,0x8f3f1f,0, false,0).setPotionName("immersiveengineering.potion.flammable");
 
-		potionID = Config.getPotionID(potionID, "Slippery");
+		potionID = getPotionID(potionID, "Slippery");
 		slippery = new IEPotion(potionID, true,0x171003,0, false,1).setPotionName("immersiveengineering.potion.slippery");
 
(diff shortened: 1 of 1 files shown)


BluSunrize:
People shouldn't configure it too high then, lol.
Besides, anyone running with Bloodmagic has a massively increased array anyway, so this is edgecase.
And lastly, I don't do 1.7 anymore.

gnif:
1) This was not configured too high, it was a vanilla out of the box generated ID running with a few other mods
2) Who cares if you don't, here is a valid working PR, if you are too lazy to run 'gradlew build' I can provide the compiled binary.

BluSunrize:
As I said: 1.7 is not getting supported anymore, get with the times. I don't endorse custom builds.
I've also replied to your necropost comment on the issue from June 2015, which you so kindly dug up for no reason.

gnif:
You yourself have made an update, build and released it on May 10th, with the last commit being one to address this exact issue, which failed to do so.

BluSunrize:
I didn't fix it, Malte did.
And it basically works, as long as people use the default values, since IE automatically tries to find free spots in the array.
Only a value configured by users to be too high causes issues.

gnif:
This WAS an auto configured value.

gnif:
Just a follow up, it doesn't try to find free spots, it assumes an id of 24 as a default and creates an instance of `IEPotion` which calls the parent constructor with the `id`... no other processing occurs. The minecraft `Potion` class takes this id and uses it as an index in the array without any bounds checking, which causes the fault. There is a function there `getNextPotionId`, but is is never used.

PR #1821, opened 2016-12-26
Add support for the Forge Energy system to Connectors

zmaster587:
Allows the LV,MV, and HV connectors to insert power into tiles implementing Forge's Energy Capability.


BluSunrize:
Well spotted, but I'd rather fix this using the utility methods I wrote for it when I added FU support to all the other stuff.

Thanks for the report though ^^

zmaster587:
Fair enough

Issue #1853, opened 2017-01-15
[DW20 1.10][IE 0.10-54] Assembler not crafting certain recipes


awebneck:
I'm attempting to use an assembler to craft a basic vanilla block of coal from 9 coal, but while it takes the recipe and shows the "ghost" output, putting coal into the assembler via any means fails to actually craft anything. I can confirm that the assembler is powered, no redstone signals are arriving at the redstone inputs, and that the problem occurs with all 3 recipe slots.

awebneck:
I should note that other recipes seem to work fine - I tried iron blocks, chests, crafting benches, etc. - all seemed to assemble except the coal block

awebneck:
Additionally, have tried entirely rebuilding multiblock and restarting both client and server, all to no avail

awebneck:
After more testing, problem does not appear to manifest in single player, but only on the server. I backed up my world and created a fresh one to see if the problem was still present, and it certainly seems to be there. Using DW20 1.10 modpack v1.3.2 with IE 0.10-54.

awebneck:
Finally got dev env setup, confirmed issue in repo master branch (only on server-side, though - single-player continues to work fine). Gonna see if I can figure it out and will issue PR.

awebneck:
Sooo... I've tracked the problem down to the recipe being returned by `Utils::findRecipe` (and therefore by `CraftingManager#getRecipeList`) seems to indicate that the recipe for coal block (as well as bale of hay and lapis block, that I could find) requires 9 items IN EACH SLOT, for a total of 81 items. Confirmed this by adding >81 coal to the assembler after putting in the recipe, and watching it go to town (and eat all 81 pieces of coal to generate a single coal block). I'm all for a challenge, but that seems a bit excessive. I'm gonna try to dig in further to see if I can figure anything else out, as this recipe appears to work fine everywhere else.

awebneck:
[...]

PR #1856, opened 2017-01-17
Only permit recipe ItemStack queries to look for a single item

awebneck:
Only permit recipe ItemStack queries to look for a single item in a stack to prevent misinterpretation of original Minecraft block recipes that call for 9 items in each recipe slot. Investigation is noted here https://github.com/BluSunrize/ImmersiveEngineering/issues/1853, but while I originally closed the issue I later wondered if there was ever any actually valid instance where a recipe calls for a stack of more than a single item in a given slot, and failed to think of any. If this is mistaken, feel free to disregard and close. However, this fixes the issue of being unable to assemble coal blocks from 9 coal that is present in the multiplayer version of the mod.


awebneck:
Closing this as the root cause was patched by Forge in https://github.com/MinecraftForge/MinecraftForge/commit/1e7eb6fb09ef6a211729e96200a203d504e10d03

Issue #1768, opened 2016-12-04
[MC 1.10.2] Crash when shooting railgun 


EtherG:
I got a crash when shooting the railgun with steel rods
Version: 0.10-48
http://paste.ee/p/3z9nw

EtherG:
I think its the railgun shot entity that's causing it.
it crashes every time i join the game

Bitterholz:
Might be wrong Import...
@BluSunrize are you sure you meant to use:
`com.google.common.base.Optional;`
instead of
`java.util.Optional;`
at
https://github.com/BluSunrize/ImmersiveEngineering/blob/master/src/main/java/blusunrize/immersiveengineering/common/entities/EntityRailgunShot.java#L7

BluSunrize:
Yes, I'm quite sure, since EntityItemFrame uses the Google optional as well ^^

I do believe that there is something messing with DataManagers, but I can't prove, nor pinpoint it.
But the fact that Botania seems to suffer a similar issue (see #1766) leads me to believe that it's not necessarily our mods that are the root cause of this.

malte0811:
@EtherG is there any difference in mod versions or configs between server and client at all? This might happen if some `DataSerializer`'s are registered only on the server or in a different order on the client vs. the server. I don't know whether any mods register serializers based on the config, but I know that there are mods registering serializers (Steve's carts and Charset register some for sure, I don't know who else)

mconwa01:
I'm getting a very similar issue with both the railgun shot and the fluorescent tube.

I see that there's some inconsistency in the first parameter to `EntityDataManager.createKey()` among the entity classes (https://github.com/BluSunrize/ImmersiveEngineering/tree/master/src/main/java/blusunrize/immersiveengineering/common/entities):
* `EntityIEProjectile` uses `EntityIEProjectile.class` (sounds right)
* `EntityChemthrowerShot` uses `EntityChemthrowerShot.class` (sounds right)
* `EntityRailgunShot`, `EntityRevolvershot` and `EntityRevolvershotFlare` all use `EntityIEProjectile.class` (sounds wrong, even though they're all subclasses of IEProjectile)
* `EntityIEExplosive` uses `EntityIEProjectile.class` (sounds wrong)
* `EntityFluorescentTube` uses `EntityChemthrowerShot.class` (sounds wrong)

Could this be where the wires are getting crossed?

[...]

PR #1886, opened 2017-02-05
Fixed EntityDataManager registration, closes #1868

### LINKED ISSUE #1868: Client crash involving fluorescent tube and revolver shot
Client and dedicated server, both with Forge 12.17.3.2185 (with MCP and FML) and IE 0.10-55. Client also has mercurius_updater 1.0. No other mods.

Crashes available at https://gist.github.com/mconwa01/cb029a6e96eb0995bf6e56ac40e23c93.

1. Create a new world on a dedicated server.
2. Connect to the server, fire a Casull Cartridge from a Revolver, and exit the client entirely (not just disconnect).
3. Restart the client, connect to the server, and place a Fluorescent Tube. The client crashes (1st crashlog).
4. Restart the client and reconnect. The client crashes a moment after connecting (2nd crashlog). This continues to happen every time the client restarts/reconnects.
5. Restart the server, restart the client, and reconnect. No crash this time. Remove the tube from the world and exit the client.
6. Restart the client and reconnect. Still no crash. Fire another Casull Cartridge and place another tube; this crashes the client again (3rd crashlog).

Placing the fluorescent tube by itself in a new world does not trigger the crash. Nor does subsequently firing a revolver shot and placing another tube. As far as I can tell, a client restart without a server restart is necessary to trigger the crash (though not sufficient; certain orderings like the one above are also necessary.)

I suspect this happens when certain classes (e.g. `EntityRevolverShot`, `EntityFluorescentTube`, as well as some other `Entity` classes in the same subpackage)  are initialized in different orders on the client and the server. More detail at https://github.com/BluSunrize/ImmersiveEngineering/issues/1768#issuecomment-272743654; this issue may be similar to/the same as #1768. If I had to guess, I'd say that each class should pass its own class object to `createKey()`.

mconwa01:
Possibly also #1768.

Best I can tell, these classes should be calling `EntityDataManater.createKey()` with their own class objects. This seems to prevent the crash.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/entities/EntityFluorescentTube.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/entities/EntityFluorescentTube.java
@@ -23,11 +23,11 @@ import net.minecraft.world.World;
 
 public class EntityFluorescentTube extends Entity implements ITeslaEntity
 {
-	private static final DataParameter<Boolean> dataMarker_active = EntityDataManager.createKey(EntityChemthrowerShot.class, DataSerializers.BOOLEAN);
-	private static final DataParameter<Float> dataMarker_r = EntityDataManager.createKey(EntityChemthrowerShot.class, DataSerializers.FLOAT);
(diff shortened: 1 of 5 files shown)


BluSunrize:
Thank you kindly. I kept forgetting about this x3

PR #1902, opened 2017-02-17
Bottling machine fixes

### LINKED ISSUE #1901: [0.10-56] CraftTweaker recipes for the Bottling Machine not consuming liquid, no matter what
It seems to detect the required amount ok, but won't consume the liquid after processing the inserted item. As an example, the line:

`mods.immersiveengineering.BottlingMachine.addRecipe(<extrautils2:wateringcan:0>, <extrautils2:wateringcan:1000>, <liquid:water> * 4000);`

takes an empty watering can from ExU2, along with 4 buckets of water, and fills it again. It won't fill until 4000mB are inserted into the machine, but once that's there it'll fill all the cans you want, never consuming the liquid.

"Vanilla" recipes (empty bottle + 1000mB of water) seem to work properly.

Tested with IE 0.10-56, ExU2 1.3.0 and Forge 2221.

malte0811:
#1901, also a dupe bug and some minor clean up (I use IntelliJ now, so I see a lot more warnings than before :smile:).


BluSunrize:
Cheers for the PR :D

PR #1986, opened 2017-03-26
Misc things

malte0811:
Some fixes and RS wire API as discussed on IRC.


malte0811:
I missed a small bit when making the RS connectors API, could you replace `TileEntityConnectorRedstone` with `IRedstoneConnector` [here](https://github.com/BluSunrize/ImmersiveEngineering/blob/master/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityConnectorRedstone.java#L139-L141)?

PR #1999, opened 2017-04-03
Add OpenComputers support to Mixer and Bottling Machine

TheGuyRunningSouth:
Added drivers and manual entries for bottling machine and mixer computer support. Tested to work on OC 1.6.2.7. 
The getTank() function for the mixer only seems to want to return the bottom fluid in the tank. I'll look more into it, but it really doesn't affect functionality that much since the mixer outputs the fluid on bottom first anyhow.


BluSunrize:
Fair enough for the typofix, although it's inappropriate in this PR =P

Full liability for this functionality is on you, because I dunno OC and can't test it ^^"

TheGuyRunningSouth:
Sounds good to me. Honestly I was kinda expecting you to either kick it back and say "put the typo back" or to remove the o yourself. 

I'm perfectly okay with the liability. I tested it myself before I made the PR and it seems stable. I've made sure notifications are enabled so I'll get the email if it breaks. Thank you for letting me contribute and for motivating me to try.

PR #2064, opened 2017-05-06
Misc fixes for build 58

malte0811:
See commit messages for details.


malte0811:
Don't merge this yet, I just noticed that I forgot a fix for a bug (wire coils not properly clearing their raycast hit data) I introduced in an earlier PR (unless I fixed it in between and forgot about it). I will fix that bug this afternoon.

BluSunrize:
roger.

malte0811:
The fix is here now, I have made one more API change that I don't think will be an issue since it is mostly technical:
I have a custom model cache for the control panels in IndustrialWires. If I want to clear that cache without 301a67fdc6e9c7571d9ca3e99390b24857a443b2 I either have to do so in code or add another command that does basically the same thing as "/cie resetrender". With this comit I can just run `IEApi.renderCacheClearers.add(PanelModel.modelCache::invalidateAll)` and have my cache cleared along with all of IE's cache.
If this is an issue I will just revert that commit.

BluSunrize:
Nah that sounds fine to me.
I'll merge this in a few hours.

PR #2083, opened 2017-05-12
Remapper&Null ItemStack fixes

malte0811:
Added a remapper as discussed on IRC yesterday. It allows worlds from MC1.10 to load in MC1.11. Also fixed a bunch of null ItemStack issues: The ones in the first commit happened when I tried to load my 1.10 dev world, the other ones where found by having IDEA show all issues related to null's and contant expressions.


BluSunrize:
snazzy!
Thanks for that! :D

PR #2096, opened 2017-05-15
Misc fixes

malte0811:
A lot of misc fixes, some related to camelCasing.
Also rewrote the TESR model renderer to include lighting. Documented since it is hard to understand otherwise (And because I haven't documented code in a long time). This is mainly used by turrets right now, they were quite bad for FPS before and looked shit without proper lighting.


BluSunrize:
I should really look into that rendering and see if I get it.
Might be useful to actually know how that works.

Cheers for this ^^

malte0811:
I also recommend reading the first few chapters of the link in one of the comments if you have the time. The linked wiki is about writing a software renderer (OpenGL running on the CPU), ir is well written and contains some things that can be useful in MC modding as well.
Unrelated, but do you know about a crash (NPE) in the return statement of the smart obj `getQuads` method? I have had that crash happen twice since I am working on 1.11, once for a turret and once for some sort of connector. I haven't managed to debug it yet, but I have set up a breakpoint to trigger on that NPE.

PR #2113, opened 2017-05-22
Blast furnace, CT's and a little MB's

malte0811:
The commit notes contain the details.
No, this does not contain the disassembling code. I'll put that into a separate PR since it should be in its own commit IMO (and you always squash PR's).


malte0811:
Warning: 725e8def6004ebcbe329df74b236a383463b3358 is a bit hacky, but not too bad IMO. Forge-patched MC completely cancels the interaction without telling the server it ever happened when `onItemUseFirst` returns anything but PASS. I work around that by manually sending the interaction packet if it wants to return anything but PASS.

BluSunrize:
Why does Forge even do this .-.

malte0811:
Probably a bug. I'll ask if I can get IRC to behave this time.

Issue #2061, opened 2017-05-04
Breaking a metal press's central piston with a Tinker's Construct hammer annihilates the piston


TomyLobo:
A metal press is 3x3x1, so I thought I'd tear it down in one piece with my 3x3x1 hammer from Tinker's Construct.
I hit the central piston with it and got everything except the piston.
I repeated it twice, same result.

EDIT: It appears the mold is gone, too.

Minecraft: 1.10.2
Forge: 12.18.3.2254
IE: 0.10-58-78
Tinker's Construct: 1.10.2-2.6.3

davqvist:
I just got a report where someone used a TC hammer to break the Coke Oven and lost one of the bricks. I could reproduce it, the same happens for the Blast Furnace. I assume this is a similar issue as the OP describes.

TomyLobo:
yep, happens for every IE multiblock I tried.

malte0811:
As a status update: Both Blu and I tried to find a fix for this a few days ago, but weren't able to find one. Block breaking code is a mess and I (probably Blu as well) don't know when this will be fixed, if ever. For now I'd say just don't break multiblocks with tools destroying multible blocks at once.

BluSunrize:
^ yeah, this, sadly. I tried for a good few hours without a solid lead on how to handle this :/

TomyLobo:
if i remember correctly, AA drill (with 3x3/5x5 upgrade) did not show this problem. I can't test that right now, though. Maybe that helps you figure out what TC is doing wrong or AA is doing right?

BluSunrize:
You're right, the AA one does it properly, and after checking the code, Ellpeck acomplishes this by breaking the middle block first, thusly disassembling the multiblock from there.

I could make this change for IE's drill, but it might tricky to convince boni of the same thing. This is why I was trying to find a solution in the block itself, because while changing my code is possible, convincing others not always is.
I'll talk to him and see what he thinks on the matter and if he'd be willing to make that change.

malte0811:
I had a new idea this morning, I will try and see whether it works. It all depends on the order in which stuff is called in breakBlock.

malte0811:
It's working. It is a bit of a weird fix though, I'll push a "demo" for just metal MB's in a minute and @BluSunrize can decide whether he likes it or not.

PR #2114, opened 2017-05-22
Breaking Multiblocks with hammers and drills

malte0811:
A fix for #2061. As discussed on IRC, it is quite hacky but since boni does not want to fix it on his end you said you would pull it.
This PR is based on #2113 to prevent merge conflicts. Please do not pull this before #2113, I intentionally split my fixes into 2 PR's to keep this in its own commit.


malte0811:
I just seriously messed up this PR by accidentally rebasing on master rather than 1.11.2, I'll clean this up and push the fix directly.

PR #2115, opened 2017-05-22
Kobalt Tweaks

KobaltCode:
Minor tweaks and game balance adjustments.
-Crusher will now 'crush' melons into melon slices
-The barrel (wooden and metal) will now show their contents regardless of what you are holding.
-Yield from crafting wires was adjusted to feel more balanced. This is also reflected on the metal press, allowing it to remain the most cost effective way of producing it.


BluSunrize:
I like the idea of melons being crushed, but the other two fixes I don't agree with.

Barrels only show contents with fluidcontainers in hand in order to not weigh down your GUI. Waila and TOP still show contents regardless, but IEs ingame overlay is specifically restricted to avoid feeling too heavy.

Currently, 4 ingots -> 4 plates -> 4 wires ->4 wire coils. Each coil has a reach of 16 blocks or more, and you're paying with a single ingot for that length of transfer. Making them any cheaper doesn't seem like a good idea.

Issue #2129, opened 2017-05-27
Crafttweaker compatibility: 1.11.2 version of IE (11-59) question


kragnoth:
I can't seem to get any of the scripts to throw an error other than no such method: immersiveengineering

Is this because there is no modtweaker for 1.11.2?  I'm not sure crafttweaker recognizes mods.immersiveengineering.

PR #2150, opened 2017-06-04
Fix CraftTweaker compat mod id. Add support for Alloy Recipes to CT

primetoxinz:
As per title.


tyra314:
Would fix #2129

BluSunrize:
Thank you kindly =3

PR #2237, opened 2017-06-27
[1.10] Adds support for Mystical Agradditions

SoftPoison:
As per title.

Do note, due to the way Mystic Agradditions handles insanium and the tier 6 inferium crop, it may break a little in future versions, but that's unavoidable

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/util/compat/IECompatModule.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/util/compat/IECompatModule.java
@@ -38,6 +38,7 @@ public abstract class IECompatModule
 		moduleClasses.put("harvestcraft", HarvestcraftHelper.class);
 		moduleClasses.put("IC2", IC2Helper.class);
 		moduleClasses.put("mysticalagriculture", MysticalAgricultureHelper.class);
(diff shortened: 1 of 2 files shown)


BluSunrize:
Is this for 1.10 or 1.11?

malte0811:
I feel like the compat should be added on Mystic Agradditions side due to requiring even more reflection than M. Agriculture when done from IE's side. In the end it is @BluSunrize's decision though. Pinging @BlakeBr0 (the author of Mystic Agriculture/Agradditions).

SoftPoison:
It's for 1.10 as the PR as that's what #2156 was merged into, but I have tested it to also be compatible with 1.11

BluSunrize:
I didn't really plan on another 1.10 release, that's why I'm asking.
I'm hoping to move towards 1.12 rather quickly this time so as not to get left behind.

SoftPoison:
Ah, fair enough. I could resubmit the PR to the 1.11 branch if you would prefer

BluSunrize:
I'm with Malte on this, develop input please, before we start reflecting *everything*

BluSunrize:
I renamed branches, which resulted in this closing. Sorry.

malte0811:
The author of Mystical Agradditions doesn't like the idea, see the above issue (#2536).

Issue #2247, opened 2017-06-28
Accept Actually Additions oil in Diesel Generator


Eufranio:
The Diesel Generator isn't accepting oil and canola oil - both fuels - from the Actually Additions mod, I suggest adding them to the fuels that the Diesel Generator can accept.

TheSpareAce:
Also make it accept Diesel fuel from Traincraft (For 1.7.10)

PR #2303, opened 2017-07-13
Improve Actually Addition's compat (Mostly Oils as Fuel)

Faxn:
* Register's Actually Addtion's fuels with the Diesel Generator, Drill and Chemthrower based on their configured fuel values in AA.  #2247
* Lets you use CanolaOil as plant oil to make BioDiesel.
* Adds AA's rice to the list of plants that you can ferment, since most anything starchy or sugary will break down into alcohol in the right conditions.

If you test this you may notice a bug with the Diesel Generator. After it has accepted one flavor of fuel, It won't accept any other kind of fuel. If you want to test them all you would need to break and remake the Generator each time. Not sure If I should report that bug yet, since It can't really show up in master, unless you have some mod that adds a second registered fuel. 

I'm not really sure this change justifies adding AA as a dependency myself. If you would prefer I could change it hard code the default fuel values so we don't need to call AA's API.


BluSunrize:
While I did kind of forget about this, I kinda don't want the AA fuels in the Diesel Gen
"Empowered Canola" is a bit too magical for my retro-industrial designs .-.

Faxn:
@BluSunrize Yeah, I see how the mods sort of clash on theme, They probably shouldn't work too well together by default.

If I made a PR that added wrappers for the "DieselHandler.registerFuel" method to crafttweaker would you be interested in that? That way it would stay true to theme by default but advanced users could make it burn whatever crazy rainbow juice some other mod maker cooks up if they so wish.

BluSunrize:
Yeah Crafttweaker compat I'm totally cool with.
If you wanna make a PR; be my guest ^^

PR #2307, opened 2017-07-14
Back to TESR wires, but better than ever!

### LINKED ISSUE #1281: [1.10] Wires do not render under certain conditions

malte0811:
*THIS IS VERY WIP, DO NOT USE OR PULL!*
I made this PR to completely fix #1281 (and I wanted to do something a bit different from what I normally do).
They use VBO's now. The performance is quite good (60FPS dropped to 45 with 1083 wires on 400 connectors). The render currently does not respect the VBO settings in MC's settings and will probably break horribly on computers that don't support VBO's. This is a PR because I want some feedback on it. I am not going to push any more commits in the next couple of weeks, but I will respond to comments on this PR. @BluSunrize you should be able to push commits to this branch if you want to.
Use `/ie genWireGrid <xMin> <y> <zMin> <xMax> <zMax>` to generate a grid of wires for performance testing. The new render only works for HV relays right now (just the model JSON needs to be changed to make the others work). All wires are rendered bright red because HV wire is hard to see against bedrock :smile:.


BluSunrize:
I've not yet toyed with it, but I'm intrigued.
I know little about VBOs, but from what I can see it seems mostly reliant on vanillla stuff with rendering, through use of the buffers.

In theory, I would imagine this to be *more* performant than the Baked one, right? Given that we're skipping the constant rebaking and caching of a shiteload of models. Instead it's just the buffers that are cached and they should be more lightweight on memory.

So performance left aside, how does it look? Is it lighting nicely? Is occlusion making it vanish, like so many TESRs?

malte0811:
It is probably a slight improvement on memory usage, but it is definitely hurting FPS (noticable only with quite large test grids (/cie genWireGrid)). Each connector has its own VBO right now which it has to render in the TESR. I may be able to improve performance by making one central VBO that is rebuilt when a connector loads/unloads on the client, lighting would be a major issue for that method (might be necessary to invalidate the VBO on any chunk update).
I don't think lighting works right now either, it should be possible to use the vanilla lighting calculations for this though. The main advantage of the VBO TESR's over models is that we can control when the wire is rendered. The TESR is set as global, so it should be rendered whenever the render AABB is visible. Occlusion should not matter unless the change to VBO's broke that (I tried fast TESR before).

BluSunrize:
Oh, so it's not too great for performance either. Well fiddlesticks, I figured that might have been one reason to do it.

So I guess the only real advantage to this is not having the partial-rendered wires anymore?

malte0811:
Yes. I might be able to significantly improve performance though, and the performance problems shouldn't be relevant with sane amounts of IE wires.

asiekierka:
@malte0811 Does Minecraft really support computers with OpenGL 1.4 or lower? OpenGL 1.5 includes VBOs, last I checked.

malte0811:
This is based on MinecraftForge/MinecraftForge#4143 (and it is quite incomplete), that PR seems to support computers without VBO's. We do enable the stencil buffer on startup, so it's possible that IE needs a newer version anyway.
This PR is currently mostly dead though, I couldn't find a good way to get notified by lighting updates (esp. in chunks with no IE blocks in them) without coremodding. There was an event for that in forge, but it was removed when models where added.
Out of couriosity, how did you come across this PR?

asiekierka:
>This PR is currently mostly dead though, I couldn't find a good way to get notified by lighting updates (esp. in chunks with no IE blocks in them) without coremodding. 

IWorldEventListener

>Out of couriosity, how did you come across this PR?

I was bored.

PR #2348, opened 2017-07-22
Fix (potential) model issues being caused by handling ItemBlocks twice

### LINKED ISSUE #2320: [Small bug] MC 1.12 missing textures
It isn't that important, but can't hurt to report it, right?

PaleoCrafter:
It seems that your adjustments for the 1.12 registry rewrite caused some unexpected model issues. Namely, the item representations of blocks are now part of `IEContent.registeredIEItems`, so `ClientProxy.registerModels` would try to register model definitions for all blocks twice. This was only noticeable ingame for the fluid block items, but might have occurred elsewhere as well. As such, this fixes #2320. I quite honestly don't understand why you're using a `ItemMeshDefinition` for the fluids, considering you could do the same with `setCustomModelResourceLocation`, but I didn't change it because you probably have your reasons and I wanted to keep the changes to a minimum.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/client/ClientProxy.java
+++ b/src/main/java/blusunrize/immersiveengineering/client/ClientProxy.java
@@ -97,6 +97,7 @@ import net.minecraft.inventory.EntityEquipmentSlot;
 import net.minecraft.item.EnumDyeColor;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemArmor;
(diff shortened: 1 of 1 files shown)


BluSunrize:
How the flip did I not think of that?
*shakes head*
Stupid me, of course, it's being registered twice.

Nicely spotted, I'll look into merging this tomorrow when I'm not overly tired. ^^

PR #2375, opened 2017-07-30
Fix UUID for my revolver!

warlordjones:
A case of mistaken identity, it seems!

### DIFF
--- a/contributorRevolvers.json
+++ b/contributorRevolvers.json
@@ -49,7 +49,7 @@
 }
 
 {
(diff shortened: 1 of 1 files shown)


BluSunrize:
So how long has that been an issue? xD

warlordjones:
Not sure. I hadn't bothered to set username in my development environment (only place I've been 'playing' lately) until now... :D

BluSunrize:
Fair!

BluSunrize:
probably should have renamed that commit. Woops.

warlordjones:
No git history I've even looked at ever remains clean for long :p

PR #2416, opened 2017-08-13
Make the coremod comply with Forge guidelines

malte0811:
This was meant to be "Make the coremod comply with the Forge guidelines", see the first commit. When I was done with that I asked Lex (the whole forge IRC channel, but he answered) whether I'd made any major mistakes. He told me a) that I did everything correctly and b) that we don't need the coremod/that moving the armor is possible without one. So I removed the coremod, as I said in the title.
This is a PR for 2 reasons:
1. This is more than a simple bugfix, I didn't want to just push it to master.
2. I left in one thing from the first commit that will require some work on your side: Signing the IE jar (I'll remove it if you don't want it). When MinecraftForge/MinecraftForge#4251 gets merged this would make it quite easy to detect versions that have been tampered with. You will have to generate a certificate for IE and put something like this in the global gradle.properties (in `<HOME>/.gradle`):
```
keyStore=...
storeAlias=...
storePass=...
```
I have no idea how to set up maven to sign jars. It should be possible, but I haven't ever worked with maven/jenkins and don't have access to a server for testing.
I added an array for keys that may be used for non-curseforge builds. My key is in there (for debug builds), the key for maven would go in there as well unless you want to use the same for maven and curseforge (which would probably be a bad idea). These keys will still log warnings, but will also tell the user that using these builds may be fine in some situations.


BluSunrize:
I chose the ASM solution for what I thought was good reason. I tried replacement before that. It was ugly and is less reliable for possible extensions and variations. 

The ASM hooks on the lowest possible layer, so that any extending classes should be covered as long as they don't fully change the render. 
I don't really want to step back to replacement...

malte0811:
Ok, so I should just revert 3972dbcbe9fb6252007b5f11163f38b1eaf86502 and turn this into the PR I originally wanted to make?
I didn't know you had tried that method before since it didn't end up on GitHub, sorry. Lex isn't going to like this.

BluSunrize:
I care reasonably little about Lex liking it. I had a longer discussion with fry on the matter. 
We agreed that forge could profit from a proper arm posing system, but I needed a speedy solution. 

I'll look into the relevant commits when I'm home. It snot great to manage this from a phone

malte0811:
Blu, did you get around to looking at the changes? I'll deal with the merge conflicts after that.
I'd also like to have the fix for #2468 in the next release so I can use it in IW, could you look over that commit as well? It isn't really messy, but it isn't perfectly nice code either, so I didn't want to push to master directly.

BluSunrize:
I hadn't really looked over this again.
I still don't really know how signing JARs works and stuff.

I'll poke Sky tonight to see if he at least knows how to make it work with Jenkins, and I'll see if I can reach Blom as well, I think he's done stuff with it before.

I'll get back to you tomorrow on this getting pulled.

BluSunrize:
Question on the changes to the build.gradle file. Am I seeing this right that it creates a separate jar for anything in the asm package? And that jar is then..."nested" in the the shipped one?

malte0811:
Yep, nested and automatically extracted. The old versions of the nested jar will be automatically overridden if the name is different IIRC. That's how coremods are supposed to work now, https://github.com/PaleoCrafter/Dependency-Extraction-Example is the only other example I could find so far.

BluSunrize:
Right, okay then, I'll give this a try, can't be too hard, and if it makes verifying legit builds easier (not that we've ever had that issue before), it's a good idea. Might as well try to be progressive here =P
Also someone linked me the Forge tutorial on it which is way easier to understand than the Oracle one.

Can you chuck a quick commit in to resolve the conflicts?

malte0811:
Done. Squash is probably the best merge option here since 4 of the 6 commits are irrelevant.

PR #2477, opened 2017-09-10
Add ore dictionary with amount ingredient support

MatrixN1:


BluSunrize:
Sure there must be a nicer way than using a Pair? Could just make it into an IE IngredientStack directly...

MatrixN1:
@BluSunrize done

BluSunrize:
Nice, thanks a lot :D

Issue #2188, opened 2017-06-18
[1.10.2] Toolbox gets "enchanted" glow when picked up


KindarConrath:
If you place an Engineer's Toolbox then pick it up by any means, it gets the enchanted glow.

Tested in Single Player

Minecraft: 1.10.2
Forge: 12.18.3.2316
Mods:
Immersive Engineering 0.10-61
Forestry 5.2.17.376
JEI 3.14.7.420
Ender IO 3.1.193
EnderCore 0.4.1.66-beta

BluSunrize:
Hahahaha, guess that's my bad.
I made it so it would keep Enchantments (in case someone used Soulbound) and now it writes empty Enchantments to the item.
It's more funny than gamebreaking, I guess...

Firedale:
Well, it IS a magical box... it can hold many things in the palm of your hand!

drewcassidy:
Honestly its a huge distraction to have it glowing in the hot bar all the time, I would really like this fixed

Carnaxus:
Also happens with the storage crates, I've got twelve not glowing and four glowing ones.

PR #2484, opened 2017-09-11
fix BluSunrize/ImmersiveEngineering#2188

hintss:
I'm not too familiar with Forge and the minecraft client, so I'm not going to say this works perfectly and won't cause any issues, however, I tested it and it seems to work, and I don't see how this would cause any unintended behaviors.


BluSunrize:
I'm not doing any further 1.10 releases, I have said this plenty of times.

PR #2652, opened 2017-12-07
Question 1.10.2

LordKhaox:
Is there anyway, a command maybe to add a mineral deposit in the chunk the player is standing in ?

Would be very need for custom map makers :)

Greetings LordKhaox


malte0811:
I assume this is a PR by accident? I put an answer to the actual question on the issue.

PR #2697, opened 2017-12-27
Added indicator for sheetmetal tank content

### LINKED ISSUE #2690: Request: TOP metalsheet tank support.
Woudl be awesome if you guys could implement TOP support for the tank. It would be neat to see the current contained fluid value.

SirWindfield:
The indicator only appears if there is actually content within the tank. I can change that behaviour however. 
I thought about making the colors dynamic and use the fluid color as fill color. But I have no idea how to get a matching text color so they have enough contrast in between. 
The other option would be to use the same style as the IF indicator.
Screenshot Screenshot of the indicator.
Resolves  #2690.

Edit: Just wanted to try dynamic colors out. Looks like `master.tank.getFluid().getFluid().getColor()` doesn't work as I expected. It always returns `0xffffffff`.


BluSunrize:
Thanks for this ^^

For future reference, to get proper readouts for colour, use getColour(FluidStack), because fluids may have their colour dependant on NBT data (like IE's potions)

SirWindfield:
Ah ok, I am quite new to modding ^^ I will try it out. Maybe the dynamic color bar looks nice. Will update with an image once I figured it out.

PR #2716, opened 2018-01-01
Show fluid amount in barrel with TOP

PomeloBooks:
I feel a little bit inconvenient to check the amount of fluid in barrel that you have to hold a bucket or something when I play. So I add it to TOP.

I also tried to get a proper fill color for the fluid as you mentioned  at #2697, but `barrel.tank.getFluid().getFluid().getColor(barrel.tank.getFluid())` seems always return `-1`.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/util/compat/OneProbeHelper.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/util/compat/OneProbeHelper.java
@@ -17,6 +17,7 @@ import blusunrize.immersiveengineering.common.blocks.IEBlockInterfaces;
 import blusunrize.immersiveengineering.common.blocks.TileEntityMultiblockPart;
 import blusunrize.immersiveengineering.common.blocks.metal.TileEntitySheetmetalTank;
 import blusunrize.immersiveengineering.common.blocks.metal.TileEntityTeslaCoil;
+import blusunrize.immersiveengineering.common.blocks.wooden.TileEntityWoodenBarrel;
 import com.google.common.base.Function;
 import mcjty.theoneprobe.Tools;
 import mcjty.theoneprobe.api.*;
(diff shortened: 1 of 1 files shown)


BluSunrize:
That would be because lava doesn't have a colour. That methid is specifically for fluids that can *change* in colour.

PomeloBooks:
I found the class `MapColor`, and now it works fine with vanilla water and lava.

lava

But not work with the creosote (haven't tested other mods ), It seems because IE uses  `Material.WATER` to [instantiate it] (https://github.com/BluSunrize/ImmersiveEngineering/blob/509f4b2f004005bda7c997399726f5c9d0708cc2/src/main/java/blusunrize/immersiveengineering/common/IEContent.java#L273).

creosote

I got two ways to solve this problem, but both need your help.
- instantiate it with a custom material
- use [this constructor](https://github.com/MinecraftForge/MinecraftForge/blob/cfbcfeaf4b5d55d5fd1af70b84139ff0d205a9b8/src/main/java/net/minecraftforge/fluids/Fluid.java#L147) to instantiate [`fluidCreosote`](https://github.com/BluSunrize/ImmersiveEngineering/blob/509f4b2f004005bda7c997399726f5c9d0708cc2/src/main/java/blusunrize/immersiveengineering/common/IEContent.java#L205)

BluSunrize:
Do other tanks have coloured TOP bars for their fluids? I don't remember. I'd like to attempt to keep it uniform to other mods in that regard.

PomeloBooks:
Actually I have only played a few of mods. After trying out some other mods I just found like Tinker Construct and Thermal Expansion also use this TOP's default color for fluid in their tank. However lava is blue (the TOP default fluid color) in Metal Barrel but 'light grey' in Fluid Tank. It seems not so uniform... Anyway color can not stop me liking Immersive Engineering.

PomeloBooks:
TOP default uses `net.minecraftforge.fluids.capability.CapabilityFluidHandler#FLUID_HANDLER_CAPABILITY` with EnumFacing of `null` to get fluid in tanks. I suggest just add the TOP's default color to Sheetmetal Tank's TOP bar, since it's not so appropriate to expose its fluid capability to null facing.

Issue #2719, opened 2018-01-04
Garden Cloche does not idle properly when its output is full


Xiaminou:
### Description of the issue:
I have been using [LagGoggles](https://minecraft.curseforge.com/projects/laggoggles) around my base to track down what might be slowing down my server and I noticed this:
When the Garden Cloche's output is full it should idle itself.

Idle:
![2018-01-04_13 39 52](https://user-images.githubusercontent.com/8670392/34565472-50c75f70-f15b-11e7-9a0b-fa38bc128ac5.png)
Processing:
![2018-01-04_13 40 18](https://user-images.githubusercontent.com/8670392/34565483-56a694ec-f15b-11e7-8788-96a41b92ddb1.png)
Output full (and crop fully grown):
![2018-01-04_13 57 40](https://user-images.githubusercontent.com/8670392/34565525-6f61356e-f15b-11e7-80c1-f26b1cdb281c.png)
![2018-01-04_14 29 58](https://user-images.githubusercontent.com/8670392/34565607-c6207e28-f15b-11e7-9f9e-bddc36e248de.png)

### Versions & Modlist
Minecraft: 1.12.2
Forge: 2583
IE: 76
Mods: [list.txt](https://github.com/BluSunrize/ImmersiveEngineering/files/1603443/list.txt)

temp1011:
I might have improved this slightly, let me take a profile

PR #2722, opened 2018-01-05
Optimisations for the Garden Cloche

temp1011:
Aiming to reduce idle differences in https://github.com/BluSunrize/ImmersiveEngineering/issues/2719. So far I've tried to reduce the code being looped through in an idle state. I realise that users can set up checks on storage to then power this with redstone and turn it off but this will hopefully reduce resources slightly.
EDIT: Added config for light animation. Grow animation probably uses more resources but I feel that's necessary for the machine.
EDIT2: I think to do this I need to rework more than the cloche itself. I might do that or just close this issue.

PS - Can I refactor this slightly? It's quite a few nested if/else loops.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityBelljar.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityBelljar.java
@@ -126,7 +126,7 @@ public class TileEntityBelljar extends TileEntityIEBase implements ITickable, ID
 				if(handler!=null&&handler.isCorrectSoil(inventory.get(1), inventory.get(0)) && fertilizerAmount>0 && energyStorage.extractEnergy(IEConfig.Machines.belljar_consumption, true)==IEConfig.Machines.belljar_consumption)
 				{
 					boolean consume = false;
-					if(growth >= 1 && !previousOutput.equals( inventory.subList(3, 7)))
+					if(growth >= 1 && !previousOutput.containsAll( inventory.subList(3, 7)))
 					{
 						ItemStack[] outputs = handler.getOutput(inventory.get(1), inventory.get(0), this);
 						int canFit = 0;


src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityBelljar.java
```diff
@@ -217,8 +222,10 @@ else if(renderActive)
 							int out = Math.min(inventory.get(j).getCount(), 16);
 							ItemStack stack = Utils.copyStackWithAmount(inventory.get(j), out);
 							stack = Utils.insertStackIntoInventory(outputTile, stack, facing);
-							if (!stack.isEmpty())
+							if (!stack.isEmpty())  {
```
malte0811: 1. Wrong codestyle, the { should be on the next line.
2. Does this work if the output is removed through the GUI? You don't seem to be resetting `outputFull` in that case.

temp1011: ah need to fix that

temp1011:
Hopefully removing from the inventory manually now works. I'm not sure about reworks, I thought the cloche's usage was quite high yesterday but in retrospect a block with a complex model and animations would do.

temp1011:
I'm not sure this will be a fruitful pull request. If it isn't feel free to close. If you have anywhere that a mediocre coder can help let me know :)

temp1011:
This may be a stupid one but won't the cloche use more fertiliser than it should due to running renderGrowth serverside? https://github.com/BluSunrize/ImmersiveEngineering/blob/aff503541904c1b79df2907fd3e1cd8cddf66f30/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityBelljar.java#L92-L102
Sorry if that's a noob question.

PR #2754, opened 2018-01-25
[BWM Compat] Add Drill to shovel toolclass to comply with HCPiles

primetoxinz:
Due to a feature in Better With Mods dirt, gravel and other soils don't drop their full block unless mined by a "shovel" toolclass. By adding the drill to this class it shouldn't change anything but make it interact with BWM correctly


BluSunrize:
Sure I guess *shrug* =P

PR #2805, opened 2018-02-21
Refactoring ItemWireCoil to make it easier for addons to extend

AntiBlueQuirk:
I'd like you to consider refactoring ItemWireCoil to separate its onItemUse function out where addons can reuse it more easily. For my addon, Alternating Flux, I had to copy this entire function to reuse it for my own wire coil, which is rather messy. Part of this is due to ItemWireCoil extending ItemIEBase, meaning addons cannot trivially extend from it to get the functionality. I also required the ability to limit which TileEntities my wire could connect to, which ItemWireCoil does not appear to support. (Connectors can choose what wires they will accept, but wires cannot choose what connectors they accept.)

This isn't a critical issue, since I've managed to get things running fine, but it would be better if this functionality was easier to access.

I've attached this request to a pull request. I don't expect this code to get accepted verbatim, but it's my idea for how this request could be fulfilled. It compiles, and doesn't appear to break any existing functionality. One note of concern is that I've added a method to the IWireCoil interface, which may break compatibility with older addons.


malte0811:
This is a good idea, thank you for the PR.
1. `doCoilUse` should probably be in `ApiUtils` or somewhere like that.
2. Why don't you make `canConnectCable` a default method? That would keep binary compatibility.
3. For Industrial Wires I use wire coils that have length stored as NBT. I'd need something more complex for that, I'll check what exactly I need after my exam.

malte0811:
I'd need a method to get the wire length for an item stack ([here](https://github.com/AntiBlueQuirk/ImmersiveEngineering/blob/0a73ff65b833e4f48996212d5bae47c670bb6958/src/main/java/blusunrize/immersiveengineering/common/items/ItemWireCoil.java#L147), defaulting to `wire.getMaxLength()`) and a method to consume the wire ([here](https://github.com/AntiBlueQuirk/ImmersiveEngineering/blob/0a73ff65b833e4f48996212d5bae47c670bb6958/src/main/java/blusunrize/immersiveengineering/common/items/ItemWireCoil.java#L200), defaulting to reducing the stacksize).

malte0811:
Blu said he wants to get a new build out tomorrow or Tuesday (mostly to fix my blast furnace derp). I have 2 things waiting for this PR to be merged to prevent merge conflicts and confusion. It'd be great if you could make the changes I requested quickly, otherwise I'll push the changes and we'll just have to deal with a few merge conflicts.

AntiBlueQuirk:
Sorry, for some reason I got the idea in my head that this wasn't going to get merged. I'll make the requested changes and fix the conflicts with master.

src/main/java/blusunrize/immersiveengineering/api/ApiUtils.java
```diff
@@ -598,7 +604,138 @@ public static WireType getWireTypeFromNBT(NBTTagCompound tag, String key)
...
+						else if(linkPos.equals(masterPos))
+							player.sendMessage(new TextComponentTranslation(Lib.CHAT_WARN+"sameConnection"));
+						else if( distanceSq > maxLengthSq)
+							player.sendMessage(new TextComponentTranslation(Lib.CHAT_WARN+"tooFar"));
+						else
+						{
+							TargetingInfo targetLink = TargetingInfo.readFromNBT(ItemNBTHelper.getTagCompound(stack, "targettingInfo"));
+							if(!(tileEntityLinkingPos instanceof IImmersiveConnectable)||
+							   !((IImmersiveConnectable) tileEntityLinkingPos).canConnectCable(wire, targetLink, offset)||
+                               !((IImmersiveConnectable) tileEntityLinkingPos).getConnectionMaster(wire, targetLink).equals(linkPos)||
```
malte0811: The formatting is a bit wrong here, but the PR looks good apart from that.

PR #2817, opened 2018-02-24
TileEntityThermoelectricGen optimization

colinrgodsey:
The TileEntityThermoelectricGen can become a major TPS slowdown when enough are present. This PR scales back calculations made in this entity to calculate temperature every 2 seconds, and outputs energy twice a second with a scaled output of 10 to make up for it. 

This has been tested as a server-side optimization for a while now (across 3 modpacks), with no issue.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityThermoelectricGen.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityThermoelectricGen.java
@@ -35,25 +35,38 @@ import javax.annotation.Nullable;
 
 public class TileEntityThermoelectricGen extends TileEntityIEBase implements ITickable, IIEInternalFluxConnector
 {
+	private int energy = 0;
+	private byte ticks = (byte)(Math.random() * 50);
+
 	@Override
 	public void update()
 	{
 		if(!world.isRemote)
 		{
-			int energy = 0;
-			for(EnumFacing fd : new EnumFacing[]{EnumFacing.DOWN,EnumFacing.NORTH,EnumFacing.WEST})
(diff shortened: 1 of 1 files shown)


src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityThermoelectricGen.java
```diff
@@ -35,25 +35,38 @@
 
 public class TileEntityThermoelectricGen extends TileEntityIEBase implements ITickable, IIEInternalFluxConnector
 {
+	private int energy = 0;
+	private byte ticks = (byte)(Math.random() * 50);
```
colinrgodsey: this is set to random to stagger calculations, this can also be just set to 0.

malte0811:
This is a good idea in principle, even though I'd like to see some proper profiling data (sampler or visualvm) on this lag first, as this is the first time I hear of it. However the implementation isn't good at all.
Pulsed energy output makes no sense here and will probably cause issues with low-tier wiring.
Calculating every two seconds also doesn't make sense. There are methods in `Block` that get called whenever a neighbour block changes, it should be enough to re-run the energy calculation whenever those methods are called.

colinrgodsey:
I'll pull some profiling data from jvisualvm after we updated to 77 on the next modpack (cant profile now because its already modified). The load normally becomes noticeable when people start placing 10s of the things near eachother. Outputing energy every 10 ticks wasn't entirely necessary, the hotspots mostly showed up in the calculations of the temperature, which really seems like it doesnt need to be computer per tick. 

I'll look into something a bit more elegant using the block change methods.

BluSunrize:
I'll hold on merging this until profiler data

BluSunrize:
Made superfluous by https://github.com/BluSunrize/ImmersiveEngineering/commit/26fcb6f8aa988ed5c023a949d8f3fc221f46ec7e

PR #2836, opened 2018-03-02
Add Beetroot Seeds to Squeezer Recipes

pap1723:
You already have the BEETROOT_SEEDS registered in your ItemStack, but they were not used anywhere. I placed them in my squeezer the other day and it didn't work.

Simple fix and I put them providing 60 mB which is between WHEAT_SEEDS and PUMPKIN_SEEDS. Feel free to change that if needed.


BluSunrize:
thanks ^^

Issue #2844, opened 2018-03-04
[0.12-78] - Crash when placing metal ladder on multiblock


topster21:
### Description of the issue:
So as the title says, whenever I place a metal ladder (from IE) on a multiblock structure from IE (be it a metal tank, refinery or a crusher) it crashes the game. 

I have tested this in both singleplayer (0.12-78) and multiplayer (0.12-77, FTB Revelations) with both the same stack trace.

### How to reproduce 

Try to place a metal ladder on the side of a completed crusher, refinery (shift right-click) or tank multiblock structure. If it doesn't crash right away, keep trying different spots on the machine. Eventually it will crash the game.

### Crashlog:
[0.12-78 Single Player](https://gist.github.com/topster21/47469a6e5f786e08f0bf783e342d717d)
[0.12-77 Multi Player FTB Revelations](https://gist.github.com/topster21/a6674fec0ba24a9050d72be8b9f543d1)

### Versions & Modlist

`Java Version: 1.8.0_162, Oracle Corporation`

#### Single Player
```
| State     | ID                   | Version      | Source                           | Signature                                |
|:--------- |:-------------------- |:------------ |:-------------------------------- |:---------------------------------------- |
| UCHIJAAAA | minecraft            | 1.12.2       | minecraft.jar                    | None                                     |
| UCHIJAAAA | mcp                  | 9.42         | minecraft.jar                    | None                                     |
| UCHIJAAAA | FML                  | 8.0.99.99    | forge-1.12.2-14.23.2.2624.jar    | e3c3d50c7c986df74c645c0ac54639741c90a557 |
| UCHIJAAAA | forge                | 14.23.2.2624 | forge-1.12.2-14.23.2.2624.jar    | e3c3d50c7c986df74c645c0ac54639741c90a557 |
| UCHIJAAAA | immersiveengineering | 0.12-78      | ImmersiveEngineering-0.12-78.jar | None                                     |

```

wBakunis:
I just tried this on singleplayer. Game crashed while placing a metal ladder on the arc furnace. I'm using IE 0.1.2 - 78 with FTB Revelations 

> The game crashed whilst unexpected error
The game crashed whilst unexpected error
Error: java.lang.ClassCastException: blusunrize.immersiveengineering.common.blocks.metal.TileEntityArcFurnace cannot be cast to blusunrize.immersiveengineering.common.blocks.metal.TileEntityLadder

[...]

Issue #2847, opened 2018-03-08
[IE 0.12-78] Subtitle Localizations


Voxel-Friend:
The diesel generator's subtitle is broken and has been for many versions. Minor annoyance at best, but thought I'd give you guys a heads up. There might be others, but haven't found any. Maybe Tesla coil?

BluSunrize:
Can you give more accurate description than "broken"?
Also, if it's been an issue for a while, begs the question why it wasn't reported sooner.

Voxel-Friend:
Sorry, here is an image to clarify
![2018-03-08_07 03 12](https://user-images.githubusercontent.com/12395262/37152395-d0953438-229e-11e8-8aa2-f5c11ce5e7ee.png)
only installed mods are:
IE 0.12-78
IP 1.1.7
Forge 14.23.2.2624

BluSunrize:
fixed by PR above

PR #2859, opened 2018-03-11
fix issues #2844 and #2847.

lgthibault:
fix issue #2844. Tested against tank, refinery, blast furnace.


lgthibault:
I'm kinda new at this git thing. Thought I was creating one pull request per fix, but somehow submitted them both. Let me know if it's a problem.

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/BlockMetalLadder.java
```diff
@@ -103,11 +104,11 @@ public AxisAlignedBB getBoundingBox(IBlockState state, IBlockAccess world, Block
 	{
 		if(this.getMetaFromState(state)==0)
 		{
-			TileEntityLadder tile = (TileEntityLadder)world.getTileEntity(pos);
-			if(tile!=null)
+			TileEntity tileEntity = world.getTileEntity(pos);
+			if(tileEntity!=null)
```
malte0811: This should probably be an `instanceof` check.

src/main/java/blusunrize/immersiveengineering/common/blocks/wooden/TileEntityWindmill.java
```diff
@@ -100,6 +100,11 @@ protected float getSpeedModifier()
 
 	public boolean checkArea()
 	{
+		//Issue #2857 prevent crashing the server when EnumFacing.rotateY() would throw IllegalStateException.
+		//That happens when facing is UP or DOWN
+		if (EnumFacing.UP.equals(facing) || EnumFacing.DOWN.equals(facing))
```
malte0811: - Could be simplified to `facing.getAxis()==EnumFacing.Axis.Y`.
 - `equals` isn't necessary here as `EnumFacing` is an enum (as it says in the name)
 - The check should probably be at some earlier point as well as here. It should never be possible to place a windmill facing up or down.
 - We generally don't put comments on fairly obvious pieces of code (Ok, we don't put comments on non-obvious code either, but we should).

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/BlockMetalLadder.java
malte0811: I'm taking about checking whether the TE extends `IEBlockInterfaces.IDirectionalTile`. If it doesn't but isn't null it will crash in line 110.

lgthibault:
Just realized my master accumulates commits. Please see corrected code in separate commits at https://github.com/BluSunrize/ImmersiveEngineering/pull/2862 and https://github.com/BluSunrize/ImmersiveEngineering/pull/2863

PR #2926, opened 2018-04-06
Networking and concurrency fixes for security and stability

JamiesWhiteShirt:
The purpose of these changes is to increase security and stability. While one can excuse ignoring events regarding security and stability for being extremely rare, consider how simple it is to avoid the events addressed in these changes. In fact, one can argue these changes also reduce complexity.

While these changes make barely any difference in Immersive Engineering itself, they are sure to prevent future issues, and will also encourage the use of good practices for those who would use this project as reference material.

To sum up the changes:
* All messages sending player names as a handle for the player no longer do so. IMessageHandlers will instead infer EntityPlayers based on the message context given the side of the IMessageHandler. This addresses a security issue where players may send these messages using the name of another player.
* All messages sending dimension IDs as a handle for the player's world no longer do so. IMessageHandlers will instead infer Worlds based on the message context given the side of the IMessageHandler. This addresses a security issue of arbitrary dimension loading where players may send these messages using another dimension ID. Note `WorldServer#getWorld(int)` is designed to load dimensions that are not loaded.
* Block positions sent in messages to the server are sanity checked using `World#isBlockLoaded(BlockPos)` before access with `World#getBlockState(BlockPos)` or `World#getTileEntity(BlockPos)`. This addresses a security issue of arbitrary chunk generation. Note these methods are designed to load chunks that are not loaded.
* All message handling work is delegated to its respective thread via the `IThreadListener` interface. This addresses numerous concurrency and race condition issues, some of which can be exploited. This solution is accepted practice to avoid all concurrent access, which the game is simply not designed to tolerate.
* Some simplification, such as serializing block positions as longs, tightening some unnecessarily wide message passing, and removing a synchronization mechanism that is no longer needed.

Here are two examples of ConcurrentModificationExceptions that may occur very rarely. These can easily be reproduced by sending absurd amounts of packets. _It should be clear that the issue arises from the concurrent modification, which is not affected by the frequency of packets._

* `MessageShaderManual.HandlerServer` vs. `IESaveData#writeToNbt`: A client may send an UNLOCK request and modify `ShaderRegistry.receivedShaders` while IESaveData is iterating it. This causes the IE save data to become an empty file. [Console trace](https://pastebin.com/1jLXycSR)
* `MessageObstructedConnection.Handler` vs. `ClientEventHandler#onRenderWorldLastEvent`: The server may send an obstructed connection while the client is iterating the collection for rendering. This crashes the client. [Crash report](https://pastebin.com/MU7vB18b)

[The testing code used to trigger these rare cases can be found here](https://github.com/JamiesWhiteShirt/ImmersiveEngineering/blob/bug-playground/src/main/java/blusunrize/immersiveengineering/common/EventHandler.java#L616).


JamiesWhiteShirt:
You can have this anyway. Sorry for trying to help out.

lgthibault:
The "security" related fixes may arguably be over the top. 

However, artificial code help raise the probability of a triggering a concurrent-related exception. They will happen with the real game too, just less often, leading to frustrating-to-reproduce issues. 

I do like the tighter network packets.

src/main/java/blusunrize/immersiveengineering/common/util/network/MessageRequestBlockUpdate.java
```diff
@@ -32,25 +31,30 @@ public MessageRequestBlockUpdate()
...
-			{
-				EventHandler.requestedBlockUpdates.offer(new ImmutablePair<Integer, BlockPos>(message.dim, message.pos));
-			}
+			WorldServer world = ctx.getServerHandler().player.getServerWorld();
+			world.addScheduledTask(() -> {
+				synchronized (EventHandler.requestedBlockUpdates)
+				{
+					if (world.isBlockLoaded(message.pos)) {
+						int dim = world.provider.getDimension();
+						EventHandler.requestedBlockUpdates.offer(new ImmutablePair<>(dim, message.pos));
```
lgthibault: Alternatively consider using a java.util.concurrent Queue, so callers need not worry about synchronized().

src/main/java/blusunrize/immersiveengineering/common/EventHandler.java
```diff
@@ -312,20 +313,14 @@ public void onWorldTick(WorldTickEvent event)
...
-			synchronized (requestedBlockUpdates)
+			while (!requestedBlockUpdates.isEmpty())
 			{
-				while (!requestedBlockUpdates.isEmpty())
+				Pair<Integer, BlockPos> curr = requestedBlockUpdates.poll();
+				World w = DimensionManager.getWorld(curr.getLeft());
+				if(w!=null)
 				{
-					Pair<Integer, BlockPos> curr = requestedBlockUpdates.poll();
-					if(FMLCommonHandler.instance().getEffectiveSide()==Side.SERVER)
```
malte0811: I don't think it's a good idea to remove the side check here? It won't cause crashes, but it will cause unnecessary block updates on the client.

src/main/java/blusunrize/immersiveengineering/common/util/network/MessageSkyhookSync.java
```diff
@@ -48,7 +49,7 @@ public void fromBytes(ByteBuf buf)
 		entityID = buf.readInt();
 		NBTTagCompound tag = ByteBufUtils.readTag(buf);
 		connection = Connection.readFromNBT(tag);
-		target = new BlockPos(buf.readInt(),buf.readInt(),buf.readInt());
+		target = BlockPos.fromLong(buf.readLong());
```
malte0811: I'd kind of like to keep compatibility with cubic chunks, which this explicitly breaks AFAIK.

src/main/java/blusunrize/immersiveengineering/common/EventHandler.java
JamiesWhiteShirt: By looking at the call hierachies I concluded that WorldTickEvent is only fired for server worlds. This may be unintentional and a bug in Forge, but Forge cannot change this without breaking assumptions like this one.

Use of `FMLCommonHandler#getEffectiveSide()` is discouraged due to the nature of how it works. In this case, where a World is available, `World#isRemote` should be used instead.

It's up to you. I can re-add the side check as a `World#isRemote` check, but it is strictly not necessary.

malte0811: I'll check whether it's called for client worlds when I'm home, if it isn't it should be fine like this.

malte0811: Yes, you're right. Good to know.

malte0811:
Thanks for the PR. @BluSunrize missed me approving it, that's the main reason this took so long.

PR #2965, opened 2018-04-25
Fix Metal Press animations and sounds with customized recipe times

codewarrior0:
Also fixes animations and sounds with stock recipes and configured metal press speed.

The original attempt in fef9ea246fca267d3b5fd95f5f55afd566f39d50 incorrectly divided by `maxTicks` in places where it didn't need to, since `fProcess` was already divided by `maxTicks`


BluSunrize:
Thanks ^^

PR #2971, opened 2018-04-28
Fix spawn interdictors (Electric Lanterns, etc) not being removed on unload

codewarrior0:
My mob spawners weren't working. I noticed they were in range of the
electric lantern and removed it. The spawners continued to not work.
I set a breakpoint in `onEntitySpawnCheck` and saw the lantern was still
in the `interdictionTiles` list, preventing mob spawns and consuming
memory.

Also removes pointless calls to `ArrayList.contains`, improving
performance; the docs for `remove` state that "If the list does not
contain the element, it is unchanged."


BluSunrize:
Thank you ^^

PR #2998, opened 2018-05-17
Fix Engineer's Workbench creating free items

codewarrior0:
If the output amount of the Blueprint recipe does not evenly divide
into 64, and there are enough inputs in the Workbench to craft more
than 64 output items, the Workbench will create free items when taking
stacks of 64 items out.

This can be observed when crafting Vacuum Tubes. Taking a stack of 64
tubes will only consume the inputs needed to craft 63 tubes, creating
one tube out of thin air.


BluSunrize:
Sorry for the late reply to this, haven't gotten around to working on IE and verifying this much.

Thanks for the fix ^_^

PR #3007, opened 2018-05-23
Update zh_tw.lang

vongola12324:
Update zh_TW i18n file.


vongola12324:
In addition, I found that there are several key in i18n files that seem to be different from the code.
For example: `ie.manual.entry.bullets_HE`
If I use `ie.manual.entry.bullets_HE` as key, the manual will not show the content, and show ie.manual.entry.bullets_HE instead. But if I use `ie.manual.entry.bullets_he`, the old key, the problem seems to be solved.

BluSunrize:
Thank you!

PR #3016, opened 2018-05-27
Fix obstructed connections not rendering or disappearing too fast on LAN server

JamiesWhiteShirt:
I believe the intended behavior is an obstructed connection created by a player should be rendered for all players.

Using the current version of IE, the following happens on a LAN server: https://streamable.com/v8zco
The obstructed connection is not rendering for the player that created it. In addition, the obstructed connection disappears twice as fast when there are two players on the LAN server.

With these changes, the following happens: https://streamable.com/xmzsm
The connection renders for both players, and disappears when it should.

This coincidentally also resolves a rare potential concurrency issue where the server thread would mutate `FAILED_CONNECTIONS` while the client is iterating over it.
I have also ensured that the client sided player tick logic for magnets and chemthrower switching only runs for the client player.

I didn't want this fix to be more invasive than necessary, but I did notice the use of AtomicInteger and a HashMap is unnecessary. I could switch to a more efficient and cleaner data structure for the obstructed connections if you would like that.


BluSunrize:
Tested and verified, thanks a lot for the fix ^_^

Sorry for the late reply.

PR #3047, opened 2018-06-10
Fixed a possible bug caused by the same cache keys for covered normal and extract belts

usernameak:


usernameak:
Sorry, closing, was wrong about that

PR #3076, opened 2018-07-03
Apply player momentum to the Chemical Thrower

TeamSpen210:
This means jumping and sprinting alters the path of the fluid, along with any other movement aids. Additionally in the air it pushes the player back slightly (by far not enough to fly around or anything, but subtly noticeable).


BluSunrize:
I understand the reason for puting player movement on the particles, smoothens it out a bit.

What prompted the added knockback however?

TeamSpen210:
It makes sense from a physics standpoint, although maybe I made it so low that there's not much point.

BluSunrize:
I'll leave it in for now. It found it a little confusing in testing, but we'll see what the community says (or if they even notice)

PR #3077, opened 2018-07-03
Improve item orientations for various items.

TeamSpen210:
Add custom position transforms for a number of items / blocks:

* Wirecutters are now held vertically in the hand.
* Conveyors are held facing forwards, or vertically on the front side for the vertical type.
  This required adding a method to `IConveyorBelt` to choose the right transformation, but it's got a default implementation.
* For uncovered metal ladders, make it look identical to the vanilla wood ladder (3D sprite).

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/api/tool/ConveyorHandler.java
+++ b/src/main/java/blusunrize/immersiveengineering/api/tool/ConveyorHandler.java
@@ -452,6 +452,21 @@ public class ConveyorHandler
 		{
 			return baseModel;
 		}
+
+		enum TransformOrient {
+			HORIZONAL,
+			VERTICAL
+		}
+
+		/**
+		 * Called to get the orientations used for the item form.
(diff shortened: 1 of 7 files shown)


BluSunrize:
Neat stuff, but ... why?
Not to devalue any work that went into this, it definitely is appreciated, but none of these things were ever brought up in issues or anything so a PullRequest out of the blue feels sorta...odd? Especially with all of these being purely cosmetic, thus subjective changes.

Any insight into what prompted this?

TeamSpen210:
I noticed these didn't have transforms, it mildly annoyed me, and I knew how to fix that so I went and did it?

BluSunrize:
Like I said, I appreciate it, I'm just confused why it wasn't brought up as an issue beforehand.
None of these renders were actually broken or anything, they just looked bad subjectively, hence why a discussion in advance, regarding possible improvements, might have been more sensible.

PR #3156, opened 2018-08-08
change the hint to the contribution highlights

mindforger:
as @thommy101 noted, the contribution guidelines are now a mere white messagebox to the right of the issue template


BluSunrize:
good call, thanks Mind

PR #3223, opened 2018-09-20
Allow Mixer to be constructed with oreSheetmetal

tgstyle:


tgstyle:
Well yes, but I actually have my own mod that does and it didn't work so I figured why not fix it =)

tgstyle:
In testing the change worked fine.

Issue #2776, opened 2018-02-11
es_MX language support


Dorzar:
Hi, I'm Lionel.
I am from Venezuela and I play Minecraft for about 5 years approximately. I have always been fascinated by the modifications that independent authors make to the game, making it much more alive and new. The downside is that most of these authors are English-speaking, which is why most of the Spanish-speaking community can not enjoy them in their entirety.

For that reason I dedicate myself to translate those mods with which I play and then I share them so that everyone else can also enjoy the mod with a Spanish speaking language. I am currently working on the translation of this fantastic mod. I have approximately 50% of it translated (it is quite extensive) and I will be posting here the updates of these translations until it is 100% so that anyone who knows the subject can help me with some sentences that could be written better or in which I have some error.

With nothing more to say, here I leave what I have,
Translation based on Minecraft es_MX

https://pastebin.com/cSRBkaBx

BluSunrize:
Please re-submit this as a pullrequest to make integration easier for us.

PR #3274, opened 2018-10-28
Readding Spanish translation

rogama25:
Somehow it got lost after 2fb66b8a4e501b43560e5b6c1c895eece3000951, ( #670 ) so i'm readding it to the project.

I have taken the translation from the file attached at #2776 , renamed to es_es instead of es_mx because it's more widely used and will keep translating the missing sentences as soon as I can.


BluSunrize:
Thanks ^^

PR #3277, opened 2018-10-30
Update ja_jp.lang

karakufire:
+ Translate not-translated entries
+ Translate titles belows "Reference to Popculture", shaders list
+ Tweak manual contents for easy hyphenation: make punctuations don't come beginning of each lines
+ Tweak some texts


BluSunrize:
Thanks a lot ^_^

PR #3296, opened 2018-11-18
Clean up references to the fermenter in SqueezerDriver

sargunv:
When connecting to a Squeezer with OpenComputers, the component currently appears as an "ie_fermenter", as if the machine is a Fermenter. This makes it difficult to tell the two machines apart when connecting to both, for example when using OpenComputers to monitor a biodiesel production system.

This PR will make it appear as an "ie_squeezer" instead.


malte0811:
Must have missed that, as you may be able to tell the squeezer driver is essentially a copy of the fermenter driver. Could you replace the other references to the fermenter in that file as well (the name of the environment, the variable name in `createEnvironment` and one of the documentation strings)? That way there won't be 2 commits for a single minor cleanup.

sargunv:
Took care of the remaining references and amended the commit + message.

BluSunrize:
Thanks for this :D

PR #3298, opened 2018-11-18
Fix inconsistencies with fluids in OC drivers

sargunv:
Currently, the format received when polling for a fluid tank state is inconsistent. The assembler, fermenter, and squeezer have one format, while other machines with internal fluid tanks have another format. 

For example, the empty state of the fermenter vs refinery: 
before-empty

And the filled state of the same machines: 
before-filled

This PR will standardize it to always output a `FluidTankInfo` object. It should remain backwards compatible with most player scripts, except for scripts that depend on oddities like the fluid name always being prefixed with "fluid."

Examples: 
after-empty
after-filled


BluSunrize:
Sure. I dunno how OC works at all, but I guess I'll put this in!

PR #3329, opened 2018-12-22
Adding an OC driver for the garden cloche / belljar

awebneck:
While many IE machines have an OpenComputers driver, the Garden Cloche / Bell Jar presently lacks one. This adds a small, read-only OpenComputers driver for same; I also added a `TileEntityBellJar#shouldGrow` method to permit the driver to easily ascertain whether or not the device is in a valid state to grow (correct soil & seed, fertilizer, power). I thought about DRY-ing up the `TileEntityBellJar#update` method to use the same one, but didn't want to fiddle with the existing implementation too much given that this is intended as a purely additive change.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityBelljar.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityBelljar.java
@@ -255,6 +255,14 @@ public class TileEntityBelljar extends TileEntityIEBase implements ITickable, ID
 		return curPlantHandler;
 	}
 
+	public boolean shouldGrow()
+	{
+		ItemStack soil = inventory.get(SLOT_SOIL);
+		ItemStack seed = inventory.get(SLOT_SEED);
+		IPlantHandler handler = getCurrentPlantHandler();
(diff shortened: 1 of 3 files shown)


src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TileEntityBelljar.java
```diff
@@ -255,6 +255,14 @@ public IPlantHandler getCurrentPlantHandler()
 		return curPlantHandler;
 	}
 
+	public boolean shouldGrow()
+	{
+		ItemStack soil = inventory.get(SLOT_SOIL);
+		ItemStack seed = inventory.get(SLOT_SEED);
+		IPlantHandler handler = getCurrentPlantHandler();
+		return (energyStorage.getEnergyStored() > IEConfig.Machines.belljar_consumption&&fertilizerAmount > 0&&handler!=null&&handler.isCorrectSoil(seed, soil)&&fertilizerAmount > 0);
```
PrincessRTFM: Shouldn't the energy check be for greater than _or equal to_ the minimum power needed for a run cycle? Doesn't really matter if the power stored drops down to zero, right? Also, why did you doubl the fertiliser check? It's the second and fifth ones.
```suggestion
		return (energyStorage.getEnergyStored() >= IEConfig.Machines.belljar_consumption&&fertilizerAmount > 0&&handler!=null&&handler.isCorrectSoil(seed, soil));
```

BluSunrize:
Waiting on response to the review.

PR #3330, opened 2018-12-24
Fix NBT mismatch caused by other mods

rinart73:
Fixed NBT mismatch caused by other mods modifying NBT data on `ItemStack` creating, but not modifying recipe `ItemStacks`.

**Example:** Astral Sorcery adds an NBT tag to all non-stackable `ItemStacks` when they're created. But apparently this happens **after** recipes were initialized. Thus `ItemStacks` in recipes don't have this new NBT data. Unfortunately AS dev doesn't want to fix this.

So if I understand correctly, the only thing that we can do is to re-create recipe ItemStacks via `.copy()`. All creation events will fire for copied ItemStacks and everything will work fine.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/api/ApiUtils.java
+++ b/src/main/java/blusunrize/immersiveengineering/api/ApiUtils.java
@@ -85,21 +85,34 @@ public class ApiUtils
 
 	public static boolean stackMatchesObject(ItemStack stack, Object o, boolean checkNBT)
 	{
-		if(o instanceof ItemStack)
-			return OreDictionary.itemMatches((ItemStack)o, stack, false)&&(!checkNBT||((ItemStack)o).getItemDamage()==OreDictionary.WILDCARD_VALUE||Utils.compareItemNBT((ItemStack)o, stack));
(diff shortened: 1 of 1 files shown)


BluSunrize:
@malte0811 I'd love your opinion on this, I would consider merging it for the release after next one (so probably some time in March/April)

JackCarterSmith:
As support for v1.12 has stopped but this patch isn't available in release 98, I have produced build [99](https://github.com/JackCarterSmith/ImmersiveEngineering/releases/tag/0.12-99) on my side for those interested. 

I just have my doubts about the ‘redistribution’ part of the licence in this specific case of discontinued support.

BluSunrize:
> As support for v1.12 has stopped but this patch isn't available in release 98, I have produced build [99](https://github.com/JackCarterSmith/ImmersiveEngineering/releases/tag/0.12-99) on my side for those interested.
> 
> I just have my doubts about the ‘redistribution’ part of the licence in this specific case of discontinued support.

You are hosting compiled binaries. That is against the license. Please remove these before we are forced to issue a takedown notice.

PR #3382, opened 2019-02-11
Reset for transferCooldown in Extracting Conveyors on Redstone High

MalkContent:
The proposed change resets the transferCooldown when the conveyor is toggled off by a redstone signal.
This makes it so the time from disabling the redstone signal to item extraction is now constant.
The previous behaviour paused the count, resulting in an uncertain "time-to-extract" which could lead to eventual over-/underflows in systems that try to extract an exact number of items by disabling the redstone signal for a timed amount, unless the timing matched perfectly with the transferTickrate.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/conveyors/ConveyorExtract.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/conveyors/ConveyorExtract.java
@@ -217,13 +217,12 @@ public class ConveyorExtract extends ConveyorBasic
 	{
 		if(!tile.getWorld().isRemote)
 		{
-			if(isPowered(tile))
+			if(this.transferCooldown > 0)
 			{
-				this.transferCooldown = 1;
+				this.transferCooldown--;
 			}
-			else if(this.transferCooldown-- <= 0)
+			if(!isPowered(tile) && this.transferCooldown <= 0)
 			{
-				this.transferCooldown = 0;
 				World world = tile.getWorld();
 				BlockPos neighbour = tile.getPos().offset(this.extractDirection);
 				if(!world.isAirBlock(neighbour))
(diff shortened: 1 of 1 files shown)


TeamSpen210:
One problem I can forsee is users exploiting this by pulsing them on and off with rapid clocks to make them extract at a faster rate, since it resets whenever you power it.

MalkContent:
That's fair.
Tbh I just winged it with the reset of transferCooldown to 1. Should you rapid pulse it, it'll extract every 2 ticks (1 tick on, 1 tick off), fastest current extract rate is 4. so i guess just up the reset value to 3 (1 tick on, 3 tick off) to match that. But since that requires you to somehow create a 1 tick redstone signal, maybe go for 2, because that'd make a repeater driven rapid pulse cycle take 4 ticks. (shortest repeater pulse is 2 ticks).
Or maybe look into how hoppers do it. I know you can reliably extract 1 item per toggle with a 4 tick rapid pulse cycle using hoppers. So far haven't bothered with trying to see if it works with rapid pulse cycles of 2 ticks. I don't even know if a pulse of 1 tick can be achieved.

MalkContent:
So I looked at hopper behaviour and it looks like it's extraction rate caps at its natural, uncontrolled rate.
I think I'll just copy that so pulsed extraction on the extraction conveyors also caps at the set extraction rate. Sound good?

MalkContent:
There. Now it just keeps decreasing the counter (stopping at 0) regardless of redstone signals, which should have the intended effect. :)

MalkContent:
Also I was tempted to just revert back to the original file and just change
`if(!tile.getWorld().isRemote&&!isPowered(tile)&&this.transferCooldown-- <= 0)`
to
`if(!tile.getWorld().isRemote&&this.transferCooldown-- <= 0&&!isPowered(tile))`
but, while not realistic, this could have transferCooldown getting into underflow :x

If it is a 100% guarantee that this can never happen, this would probably be the most elegant way to handle this, but I don't know enough about MC to make that call.

PR #3404, opened 2019-03-03
Fix #3403: Add proxying block for Diagonal Downward Conveyors

### LINKED ISSUE #3403: Diagonal Downward Conveyor Insertion
### Description of the issue:
Diagonal conveyors only occupy the block they're in, so devices that should logically be able to insert onto them from diagonally above cannot, whilst adjacent devices are able to insert items "through" the surface of the conveyor. This doesn't affect upward conveyors since they can detect and insert into diagonal blocks. This affects IE machines as well, though they could be special-cased of course.

conv_bug

My suggested fix is to make these downward conveyors spawn a air-like proxy block above themselves if the space is clear, which relays the `insertItem()` call down to the conveyor below. If a block is placed above, the conveyor would effectively be blocked by that. Direct insertion would then be disallowed on the side of the actual conveyor block. If this seems reasonable I'll be submitting a PR to implement this.

### Versions & Modlist
IE: 89 / 011c70c8929dc965645dd073df451ea58d29ae60
Forge: 14.23.5.2768

TeamSpen210:


BluSunrize:
See related issue, I have no intention to add this feature.

PR #3416, opened 2019-03-14
fix: Fix speelling mistake

FledgeXu:
There are some spelling mistakes, This PR fix it.

change `registerMagnetSupression` to `registerMagnetSuppression`
change `applyMagnetSupression` to `applyMagnetSuppression`
change `revertMagnetSupression` to `revertMagnetSuppression`

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/api/tool/ConveyorHandler.java
+++ b/src/main/java/blusunrize/immersiveengineering/api/tool/ConveyorHandler.java
@@ -131,7 +131,7 @@ public class ConveyorHandler
 	 * the reversal function is optional, to revert possible NBT changes
 	 * the tileentity parsed is an instanceof
 	 */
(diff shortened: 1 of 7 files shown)


malte0811:
We can't really do that in the middle of an MC version. The functions you changed are (officially) part of the API, so other mods may be using them. If we change them now those mods will crash with IE build 90. Might be a good idea to fix them as part of the 1.13 update though.

FledgeXu:
I notice that. 
I agree that it should be fixed in 1.13 update.

FledgeXu:
Maybe we should add `@Deprecated ` to original function and than

FledgeXu:
Or maybe we should add `@Deprecated` to original function and then just call new function in original function?

BluSunrize:
Closing this, since these issues should be addressed in 1.14 or other future versions

PR #3420, opened 2019-03-18
Added crusher NBT tag 'Active'

trygve55:
The NBT tag "active" has been added to the Fermenter multiblock machine. This is for comparatility with other mods like Pollution of the Realms.


BluSunrize:
That's not how you do compatability. You don't use the NBT of another block to implement compat. That is bad style, and I won't pull that.

I do believe I have explained ths before as well, I just can't find it.

PR #3440, opened 2019-04-11
Option to disable Multiblocks and change speed of Skyline Hook

Kurtchekov:
Added following options to config file:
- enable/disable multiblock structures individually
- adjust speed values for Skyline Hooks


TeamSpen210:
For the multiblocks, this is implemented already to an extent. Setting NBT on the Hammer (by editing the recipe for example) allows you to blacklist/whitelist what multiblocks can be constructed.

BluSunrize:
I have explained previously that I have no intention to allow the removal or disabling of Multiblocks from the mod

PR #3481, opened 2019-05-29
Set up Travis CI

aDotInTheVoid:
You may need to link your GitHub to Travis [here](https://travis-ci.com/signin) for this to work

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,2 @@
+sudo: false
+language: java


malte0811:
What would be the advantage of this over the existing build server? As I understand it that would be all we'd use Travis for.

aDotInTheVoid:
Oh, I didn't know you had a build server.

*But* Travis plays really nicely with GitHub and can show CI status for each commit as well as showing build status for pull requests.

BluSunrize:
Yeah we have an existing build server already, and IE is probably never getting unit tests, so there is zero reason to have Travis.
(Plus, out of politeness, it'd be best to approach us about stuff like this beforehand >_>)

PR #3529, opened 2019-07-27
IEStairsBlock ported. Review required.

stfwi:
Java implementation ported similar to the fence implementation. Not sure if the flammability is still needed, so left in.


src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
```diff
@@ -132,13 +134,19 @@ public boolean canRenderInLayer(BlockState state, BlockRenderLayer layer)
...
 
 	@Override
-	public int getLightValue(BlockState state, IWorldReader world, BlockPos pos)
+	public int getLightValue(BlockState state)
 	{
+		//TODO double check. getLightValue() is a more tricky one now (at least at the present time).
+		//  	 Light values are defined in the block properties at construction, and it seems that
+		//	 	 values returned here must be 0 or exactly match that light value. Vanilla blocks with
+		//		 switched light use something like: `return LIT ? super.getLightValue(state) : 0`.
+		//     -> So it might be good to drop `lightOpacity` and use the vanilla frame here.
```
malte0811: Isn't `lightOpacity` something different from `lightValue`? The first one changes how much light is blocked (glass vs wood), the second one how much is emitted (torch vs RS torch).

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
```diff
@@ -174,15 +184,17 @@ public boolean isOpaqueCube(IBlockState state)
...
 	{
 		return !notNormalBlock;
 	}
 
 	@Override
+	@SuppressWarnings("deprecation")
 	public boolean isNormalCube(BlockState state, IBlockReader world, BlockPos pos)
 	{
-		return notNormalBlock;
+		return notNormalBlock; //TODO review: should this be `!notNormalBlock`?
```
malte0811: You're right, looks like a copy-paste error.

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
```diff
@@ -279,19 +293,19 @@ public ResourceLocation createRegistryName()
 	}
 
 	@Override
+	@SuppressWarnings("deprecation")
 	public boolean onBlockActivated(BlockState state, World world, BlockPos pos, PlayerEntity player, Hand hand,
-									Direction side, float hitX, float hitY, float hitZ)
+																	BlockRayTraceResult hit)
```
malte0811: The indenting is strange here, what do you use for formatting? There's an IntelliJ formatting XML for IE available [here](https://gist.github.com/BluSunrize/5c05af8f29de9581426551d9b29b1809).

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
```diff
@@ -302,33 +316,29 @@ public IELadderBlock(String name, Block.Properties material,
 			super(name, material, itemBlock, additionalProperties);
 		}
 
+		//TODO review: actually not sure if this override is still needed
+		// 			  	It appears to handle the moment when players jump
+		//				  on a ladder, preventing fall damage, and limit
+		//				  the motion accordingly. This should be already
+		//				  the vanilla default by now.
```
malte0811: I think you're right, we probably need to override something like `isLadder` then. `IELadderBlock` can probably be removed in that case.

src/main/java/blusunrize/immersiveengineering/common/blocks/IEMultiblockBlock.java
```diff
@@ -63,8 +67,11 @@ public void onReplaced(BlockState state, World world, BlockPos pos, BlockState n
 	}
 
 	@Override
-	public void getDrops(BlockState state, NonNullList<ItemStack> drops, World world, BlockPos pos, int fortune)
+	@SuppressWarnings("deprecation")
+	public List<ItemStack> getDrops(BlockState state, LootContext.Builder builder)
 	{
+		//TODO review: Presuming the drop list has to be empty, as items are manually spawned in `onReplaced()` above.
```
malte0811: Yes. It might be possible to move spawning here with the loot context, but I don't think it would be useful in this case.

src/main/java/blusunrize/immersiveengineering/common/blocks/IEStairsBlock.java
```diff
@@ -9,36 +9,40 @@
...
-//		ImmersiveEngineering.registerBlock(this, ItemBlockIEStairs.class, name);
+		super(state, properties);
+		setRegistryName(new ResourceLocation(ImmersiveEngineering.MODID, name));
 		IEContent.registeredIEBlocks.add(this);
-		IEContent.registeredIEItems.add(new ItemBlockIEStairs(this));
+		Item.Properties itemProps = new Item.Properties().group(ImmersiveEngineering.itemGroup);
+		IEContent.registeredIEItems.add(new BlockItem(this, itemProps));
 	}
 
+	//TODO: review required if this is needed.
```
malte0811: It doesn't seem to be used in 1.12, so feel free to remove it.

src/main/java/blusunrize/immersiveengineering/common/blocks/IEStairsBlock.java
```diff
@@ -51,36 +55,16 @@ public IEStairsBlock setHasFlavour(boolean hasFlavour)
...
-		return super.doesSideBlockRendering(state, world, pos, face);
-	}
-
-	public IEStairsBlock setRenderLayer(BlockRenderLayer layer)
-	{
-		this.renderLayer = layer;
-		return this;
-	}
-
+	//TODO: review: added flavour information here, so that `ItemBlockIEStairs` can be removed.
```
malte0811: 1. You're right, that class can probably be removed now.
2. The implementation seems overly complicated. The method shouldn't be called for anything but the `this`-block, so it could be something like
```java
if (hasFlavour)
tooltip.add(new TranslationTextComponent(Lib.DESC_FLAVOUR+getRegistryName().getPath()));
```

src/main/java/blusunrize/immersiveengineering/common/blocks/IETileProviderBlock.java
```diff
@@ -468,6 +469,8 @@ public int getRenderColour(BlockState state, @Nullable IBlockReader worldIn, @Nu
 		return 0xffffff;
 	}
 
+	//TODO BlockFaceShape Gone, commented out for main author deletion.
```
malte0811: `IFaceShape` seems to be the only "interesting" case, and I'd guess that MC can figure that out by itself as well now.

src/main/java/blusunrize/immersiveengineering/common/blocks/ItemBlockIESlabs.java
```diff
@@ -32,24 +17,31 @@ public ItemBlockIESlabs(Block b)
 		super(b);
 	}
 
+	//TODO: review: BlockIESlab is now a `SlabBlock`, `IESlabTileEntity` gone
+	// 			(due to flattening I suppose).
+	//      Vanilla onItemUse will invoke tryPlace, which referrs to the Block,
+	//			which is the SlabBlock. So snow layer replacements etc should be
+	//		  handled.
+	//   -> Proposal would be: testing first without overridden `onItemUse`.
```
malte0811: I'd suggest completely removing this class, IIRC it's only necessary because we used the same block for multiple types of slab.

src/main/java/blusunrize/immersiveengineering/common/blocks/ItemBlockIEStairs.java
```diff
@@ -8,26 +8,33 @@
...
 import net.minecraft.world.World;
-
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.api.distmarker.OnlyIn;
 import javax.annotation.Nullable;
 import java.util.List;
+*/
 
+//TODO This class should be obsolste by now. The tooltip can be handled in `IEStairsBlock`.
+// 		 Commented out for main author review/removal.
```
malte0811: Same as for `ItemBlockIESlabs`.

src/main/java/blusunrize/immersiveengineering/common/blocks/plant/HempBlock.java
```diff
@@ -107,22 +110,35 @@ public EnumPlantType getPlantType(IBlockReader world, BlockPos pos)
...
 	}
 
+	//TODO: Could not see a proper way to access a fortune values using the builder (e.g. luck scaling is -1024..1024)
+	//      -> added a concept loot table file data/immersiveengineering/loot_tables/blocks/hemp.json.
+	// 				 for experimenting with alternatives to overriding `getDrops()`.
 	@Override
-	public void getDrops(BlockState state, NonNullList<ItemStack> drops, World world, BlockPos pos, int fortune)
+	public List<ItemStack> getDrops(BlockState state, LootContext.Builder builder)
 	{
+		final int fortune = 0;
```
malte0811: Can't you get the fortune value from the loot context?
Edit: Isn't this whole method unnecessary now that you've written the loot table JSON?

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
stfwi: Yes, I shuffled this. Pushed commit with the `getOpacity()` override.

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
stfwi: Changed in f0088b34.

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
stfwi: Thanks for the hint, imported the XML and corrected the indent (f0088b34).

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
stfwi: Added `isLadder()` override, also added notes in the wood/metal ladder blocks, which may be better inherited directly from `mc::LadderBlock`. I am not entirely sure yet because I don't know what additional meaning these ladders have when forming multi blocks.

src/main/java/blusunrize/immersiveengineering/common/blocks/IEMultiblockBlock.java
stfwi: Removed the comment accordingly.

src/main/java/blusunrize/immersiveengineering/common/blocks/IEStairsBlock.java
stfwi: Setter removed accordingly.

src/main/java/blusunrize/immersiveengineering/common/blocks/IEStairsBlock.java
stfwi: Aye, that was a bit overdone. Applied your suggested code.

src/main/java/blusunrize/immersiveengineering/common/blocks/IETileProviderBlock.java
stfwi: Ok, I left the commented method in the code, with the removal marker for the porting-cleanup.

src/main/java/blusunrize/immersiveengineering/common/blocks/ItemBlockIESlabs.java
stfwi: Also here, marked for removal.

src/main/java/blusunrize/immersiveengineering/common/blocks/ItemBlockIEStairs.java
stfwi: Same here, I also think it's not needed anymore.

src/main/java/blusunrize/immersiveengineering/common/blocks/plant/HempBlock.java
stfwi: I did not find out how. Traced the functionality down to variable/variant property parsing from JSONs. There is a luck value that is a builder input. When you actually want to build you have to specify a loot param set, which as checked required and optional properties. This maps to types, conditions, etc. It actually throws if something is not right (or I am missing an easy bypass like "default empty param set" or so.
So that was the reason why I concept JSON file. It does not have exactly the drop probabilities as the 1.12 method. If this is okay, then we probably should go for the json file.

stfwi:
Review changes applied.

src/main/java/blusunrize/immersiveengineering/common/blocks/plant/HempBlock.java
malte0811: This looks like it should work (and does compile):
```java
ItemStack tool = builder.get(LootParameters.TOOL);
int looting = 0;
if(tool!=null)
	looting = EnchantmentHelper.getEnchantmentLevel(Enchantments.FORTUNE, tool);
```
Most of the parameters seem to be set before `getDrops` is called, (hopefully) including `LootParameters.TOOL`.

malte0811:
If you delete the methods you marked for "deletion by main authors" and remove the `TODO`-comments that are now unnecessary this should be ready to be merged.

src/main/java/blusunrize/immersiveengineering/common/blocks/plant/HempBlock.java
stfwi: Ok, I would actually prefer this solution over a loot table JSON file, so let's drop the JSON.

PR #3541, opened 2019-08-20
1.14 port: assets [1]

stfwi:
Partial asset porting. We need to synchronize about how the flattened regnames will be. E.g. "item_router" instead of "wooden_device_xyz" etc. 
Because the item models have to be explicitly listed now, a parent block model file should exist for each block.
Yet un-ported Forge blockstate files with variants need some additional attention, too - they contain data that should be now in model files.


malte0811:
I'm not going to manually check these changes, I'll just hope they mostly work and sort out the broken parts once everything compiles.

The registry names etc. should no longer have the prefixes from 1.12, as those don't make much sense any more.

PR #3573, opened 2019-10-02
[1.14] Manual: Added missing categories

SuperSecretSettings:
This might help reduce conflicts while JStocke and I work on the manual.

<details><summary>Construction Cat</summary></details>


malte0811:
How did you choose the weights? If it's just to mimic the order in 1.12: General should stay on top, but the order of the other ones doesn't really matter IMO, so you should just set all their weights to the same number (they'll be sorted alphabetically in that case).

SuperSecretSettings:
Oh ok, I’ll make sure to do that when I get home.

src/main/java/blusunrize/immersiveengineering/client/ClientProxy.java
```diff
@@ -1367,4 +1375,4 @@ public void startSkyhookSound(SkylineHookEntity hook)
 		Minecraft.getInstance().getSoundHandler().play(new SkyhookSound(hook,
 				new ResourceLocation(MODID, "skyhook")));
 	}
-}
\ No newline at end of file
+}
```
SuperSecretSettings: Rip I blame GitHub for this newline

PR #3605, opened 2019-11-25
Special recipes1.14

MalkContent:
added class with event listener to register custom recipes

redid speedloader filling recipe
ported potion bullet fill recipe
ported flare bullet color recipe

some minor code changes here and there
also autoformatted all the things :| my bad


malte0811:
> autoformatted all the things

You're using IntelliJ, right? Set it to use [this](https://gist.github.com/BluSunrize/5c05af8f29de9581426551d9b29b1809) formatting style, it's what's used for everything else in IE.

MalkContent:
done. I will say that while the open-curly-bracket-goes-to-new-line disagrees with me a little, the lack of whitespace around operators hurts my soul. That will be the first and last time I complain, though, just needed it out of the system :D

In other news: I invite comments regarding the RecipeSerializer. For one, I moved the recipe's serializers there, which were previously located as a field in the respective recipe - and I have to admit, I only did that in my confusion as to how to get them to register and it's not necessary. And then there's the event listener registering all the fields using reflection, and that was one of these "bright" ideas I had in the middle of the nights, and I now heavily question it's elegance. But hey, it works, so I won't roll it back until someone more qualified has something to say :D

src/main/java/blusunrize/immersiveengineering/common/crafting/PotionBulletRecipe.java
```diff
@@ -8,52 +8,57 @@
...
+			{
+				if(isPotionBullet(stackInSlot))
+				{
+					if(hasBullet)
+						return false;
+					hasBullet = true;
+				}
+				else if(stackInSlot.getItem() instanceof PotionItem)
+				{
+					if(hasBullet)
```
malte0811: That should probably be `hasPotion`.

src/main/java/blusunrize/immersiveengineering/common/crafting/SpeedloaderRecipe.java
```diff
@@ -8,69 +8,130 @@
...
+					if(!hasSpeedloader)
+						blindBulletGrid[i/width][i%width] = true;
+					else if(Math.abs((i/width)-speedloaderY) > 1||Math.abs((i%width)-speedloaderX) > 1)
+						return false;
+				}
+				else
+					return false;
+			}
+		}
+		return hasSpeedloader&&hasBullets;
```
malte0811: This implementation seems unnecessarily complicated. I'd suggest first finding the speedloader and then checking the bullets. Yes, it will (theoretically) be slower, but there's only one of these recipes, so performance should not be an issue.

src/main/java/blusunrize/immersiveengineering/common/data/RecipeSerializers.java
```diff
@@ -0,0 +1,41 @@
...
+public class RecipeSerializers
+{
+	public static final SpecialRecipeSerializer<SpeedloaderRecipe> SPEEDLOADER_LOAD = IRecipeSerializer.register(ImmersiveEngineering.MODID+":crafting_special_speedloader_load", new SpecialRecipeSerializer<>(SpeedloaderRecipe::new));
+	public static final SpecialRecipeSerializer<PotionBulletRecipe> POTION_BULLET_FILL = IRecipeSerializer.register(ImmersiveEngineering.MODID+":crafting_special_potion_bullet_fill", new SpecialRecipeSerializer<>(PotionBulletRecipe::new));
+	public static final SpecialRecipeSerializer<FlareBulletRecipe> FLARE_BULLET_COLOR = IRecipeSerializer.register(ImmersiveEngineering.MODID+":crafting_special_flare_bullet_color", new SpecialRecipeSerializer<>(FlareBulletRecipe::new));
+
+	@SubscribeEvent
+	public static void register(Register<IRecipeSerializer<?>> event)
+	{
+		Field[] fields = RecipeSerializers.class.getDeclaredFields();
```
malte0811: Please change this to not use reflection. Forge has `DeferredRegister` (see Lex's mods for examples), that should make this doable without much code duplication.

src/main/java/blusunrize/immersiveengineering/common/crafting/PotionBulletRecipe.java
MalkContent: yep, done

src/main/java/blusunrize/immersiveengineering/common/data/RecipeSerializers.java
MalkContent: done, but please double check that I did that correctly.
I mean everything appears to be working, but still

src/main/java/blusunrize/immersiveengineering/common/crafting/SpeedloaderRecipe.java
MalkContent: replaced it with a more necessarily complicated check that now allows for adding bullets to a non-empty speedloader :3

MalkContent:
I did add the jerrycan back in on the go.  Wanted to do that in an extra branch, but because I only added it to be able to add it's special recipe, it found its way in here. Didn't get around to the special recipe yet, but anyways, that's why that's in there.

src/main/java/blusunrize/immersiveengineering/common/data/Recipes.java
```diff
@@ -41,14 +46,20 @@
...
 	private final HashMap<String, Integer> PATH_COUNT = new HashMap<>();
 
 	private final int standardSmeltingTime = 200;
 	private final int blastDivider = 2;
 
+	public static final DeferredRegister<IRecipeSerializer<?>> RECIPE_SERIALIZERS = new DeferredRegister<>(ForgeRegistries.RECIPE_SERIALIZERS, ImmersiveEngineering.MODID);
+
+	public static final RegistryObject<IRecipeSerializer<?>> SPEEDLOADER_LOAD = RECIPE_SERIALIZERS.register("crafting_special_speedloader_load", () -> new SpecialRecipeSerializer<>(SpeedloaderRecipe::new));
+	public static final RegistryObject<IRecipeSerializer<?>> POTION_BULLET_FILL = RECIPE_SERIALIZERS.register("crafting_special_potion_bullet_fill", () -> new SpecialRecipeSerializer<>(PotionBulletRecipe::new));
+	public static final RegistryObject<IRecipeSerializer<?>> FLARE_BULLET_COLOR = RECIPE_SERIALIZERS.register("crafting_special_flare_bullet_color", () -> new SpecialRecipeSerializer<>(FlareBulletRecipe::new));
```
malte0811: Please move this somewhere else, I don't want the `data`-package to do anything in non-datagenerator runs. Similar to how the `client`-package shouldn't be touched on a dedicated server. Also, why are these `IRecipeSerializer<?>`'s rather than a more specific generic? Is that simply a limitation of `DeferredRegister`?

MalkContent: didn't chose a more specific generic, because I didn't think it mattered, because the only thing potentially regularly calling it is IRecipeSerializer getSerializer(). changed that.

as for moving the stuff: a good idea where to?
I have to keep these 3 lines together with `RECIPE_SERIALIZERS`, otherwise I can't simply keep them static final and have to figure out how to properly initialize them across different classes.

PR #3616, opened 2019-12-14
Lots of Recipes

TwistedGate:
Contains recipes for:
Voltmeter, capacitor-backpack, toolbox, Engineer's Workbench, treated scaffold, metal scaffolds (and their variations), both reinforced and normal crate, all fences, engineering blocks (redstone, light, heavy), all coil blocks, generator and finaly radiator.

Edit: And also slight formatting for the already existing recipes aswell.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/data/Recipes.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/data/Recipes.java
@@ -84,14 +84,18 @@ public class Recipes extends RecipeProvider
 			}
 			addStandardSmeltingBlastingRecipe(dust, ingot, 0, out, "_from_dust");
 //			addStandardSmeltingBlastingRecipe(dust, ingot, metal.smeltingXP, out, "_from_dust"); //TODO: remove this, if 0 XP on dust is intentional. this bugs out because the alloys do not have metal.smeltingXP
-			ShapelessRecipeBuilder.shapelessRecipe(plate).addIngredient(IETags.getTagsFor(metal).ingot).addIngredient(Tools.hammer).addCriterion("has_"+metal.tagName()+"_ingot", this.hasItem(IETags.getTagsFor(metal).ingot)).build(out, toRL("plate_"+metal.tagName()+"_hammering"));
+			ShapelessRecipeBuilder.shapelessRecipe(plate)
+				.addIngredient(IETags.getTagsFor(metal).ingot)
+				.addIngredient(Tools.hammer)
+				.addCriterion("has_"+metal.tagName()+"_ingot", this.hasItem(IETags.getTagsFor(metal).ingot))
+				.build(out, toRL("plate_"+metal.tagName()+"_hammering"));
 			ShapedRecipeBuilder.shapedRecipe(sheetMetal, 4)
-					.key('p', plate)
-					.patternLine(" p ")
-					.patternLine("p p")
-					.patternLine(" p ")
-					.addCriterion("has_"+toPath(plate), hasItem(plate))
-					.build(out);
+				.key('p', plate)
+				.patternLine(" p ")
+				.patternLine("p p")
+				.patternLine(" p ")
+				.addCriterion("has_"+toPath(plate), hasItem(plate))
+				.build(out);
 		}
 		addStandardSmeltingBlastingRecipe(IEItems.Ingredients.dustHopGraphite, Ingredients.ingotHopGraphite, 0.5F, out);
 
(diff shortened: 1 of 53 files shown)


src/main/java/blusunrize/immersiveengineering/common/data/Recipes.java
```diff
@@ -107,77 +112,556 @@ protected void registerRecipes(@Nonnull Consumer<IFinishedRecipe> out)
...
+			.key('s', IETags.treatedStick)
+			.addCriterion("has_treated_planks", hasItem(IETags.getItemTag(IETags.treatedWood)))
+			.build(out);
+		ShapedRecipeBuilder.shapedRecipe(MetalDecoration.aluScaffolding.get(MetalScaffoldingType.STANDARD), 6)
+			.patternLine("iii")
+			.patternLine(" s ")
+			.patternLine("s s")
+			.key('i', IETags.getTagsFor(EnumMetals.ALUMINUM).ingot)
+			.key('s', IETags.aluminumRod)
+			.addCriterion("has_"+toPath(MetalDecoration.aluScaffolding.get(MetalScaffoldingType.STANDARD)), hasItem(MetalDecoration.aluScaffolding.get(MetalScaffoldingType.STANDARD)))
```
malte0811: Shouldn't this be 2 calls to `hasCriterion` for alu ingots and rods? Unlocking recipes with the output item seems strange.

src/main/java/blusunrize/immersiveengineering/common/data/Recipes.java
```diff
@@ -107,77 +112,556 @@ protected void registerRecipes(@Nonnull Consumer<IFinishedRecipe> out)
...
+			.addCriterion("has_"+toPath(MetalDecoration.steelScaffolding.get(MetalScaffoldingType.STANDARD)), hasItem(MetalDecoration.steelScaffolding.get(MetalScaffoldingType.STANDARD)))
+			.build(out, toRL("steel_scaffolding_grate_top_from_standard"));
+		ShapelessRecipeBuilder.shapelessRecipe(MetalDecoration.steelScaffolding.get(MetalScaffoldingType.WOODEN_TOP))
+			.addIngredient(MetalDecoration.steelScaffolding.get(MetalScaffoldingType.GRATE_TOP))
+			.addCriterion("has_"+toPath(MetalDecoration.steelScaffolding.get(MetalScaffoldingType.GRATE_TOP)), hasItem(MetalDecoration.steelScaffolding.get(MetalScaffoldingType.GRATE_TOP)))
+			.build(out, toRL("steel_scaffolding_wooden_top_from_grate_top"));
+		ShapelessRecipeBuilder.shapelessRecipe(MetalDecoration.steelScaffolding.get(MetalScaffoldingType.STANDARD))
+			.addIngredient(MetalDecoration.steelScaffolding.get(MetalScaffoldingType.WOODEN_TOP))
+			.addCriterion("has_"+toPath(MetalDecoration.steelScaffolding.get(MetalScaffoldingType.STANDARD)), hasItem(MetalDecoration.steelScaffolding.get(MetalScaffoldingType.STANDARD)))
+			.build(out, toRL("steel_scaffolding_standard_from_wooden_top"));
```
malte0811: This block should probably be turned into a loop along the lines of
```java
int numScaffoldingTypes = MetalScaffoldingTypes.values().length;
for (MetalScaffoldingType from : MetalScaffoldingType.values()) {
    MetalScaffoldingType to = MetalScaffoldingType.values()[(from.ordinal()+1)%numScaffoldingTypes);
   //Add conversion recipes from `from` to `to`
}
```

src/main/java/blusunrize/immersiveengineering/common/data/Recipes.java
```diff
@@ -107,77 +112,556 @@ protected void registerRecipes(@Nonnull Consumer<IFinishedRecipe> out)
...
 
-		ShapedRecipeBuilder.shapedRecipe(StoneDecoration.alloybrick, 2).patternLine("sb").patternLine("bs").key('s', Tags.Items.SANDSTONE).key('b', Tags.Items.INGOTS_BRICK).addCriterion("has_brick", hasItem(Tags.Items.INGOTS_BRICK)).build(out);
+		ShapedRecipeBuilder.shapedRecipe(WoodenDecoration.treatedFence, 3)
+			.patternLine("isi")
+			.patternLine("isi")
+			.key('i', IETags.getItemTag(IETags.treatedWood))
+			.key('s', IETags.treatedStick)
+			.addCriterion("has_treated_planks", hasItem(IETags.getItemTag(IETags.treatedWood)))
```
malte0811: Maybe add a second criterion on `has_treated_sticks`? (same for other recipes)

src/main/java/blusunrize/immersiveengineering/common/data/Recipes.java
TwistedGate: Oh whops, meant to change it befor commiting but forgot. Gonna work on that now.

TwistedGate:
Anything else that would need to be done?

src/main/java/blusunrize/immersiveengineering/common/data/Recipes.java
```diff
@@ -107,77 +112,581 @@ protected void registerRecipes(@Nonnull Consumer<IFinishedRecipe> out)
...
+			MetalScaffoldingType from = MetalScaffoldingType.values()[i];
+			MetalScaffoldingType to = null;
+			if((i+1)>=MetalScaffoldingType.values().length) // Meant to prevent this whole thing to throw IndexOutOfBounds.
+			{
+				to = MetalScaffoldingType.STANDARD; // And to have wood_top -> standard conversion
+			}
+			else
+			{
+				to = MetalScaffoldingType.values()[i+1];
+			}
```
malte0811: What I suggested in my comment shouldn't throw AIOOBEs, that's what the `%numScaffoldingTypes` is for. If you want to keep things this way you should remove the `{}`s for the if, in IE we don't use them for statements with a single-line body (I don't really agree with that style in general, but this feels like far too "big" vertically).

PR #3623, opened 2019-12-26
Refactor - Shader Registry

BluSunrize:
Registration of shaders and shader cases is now done with ResourceLocations instead of Strings.

I briefly considered using a ForgeRegistry instead. Do you think that's worth it? If so I can still do that before we merge.


src/main/java/blusunrize/immersiveengineering/api/shader/ShaderCaseDrill.java
```diff
@@ -27,9 +28,9 @@ public ShaderCaseDrill(Collection<ShaderLayer> layers)
 	}
 
 	@Override
-	public String getShaderType()
+	public ResourceLocation getShaderType()
 	{
-		return "immersiveengineering:drill";
+		return new ResourceLocation("immersiveengineering", "drill");
```
malte0811: I'd prefer `ImmersiveEngineering.MODID`, that should eliminate any possibility of typos. And the compiler should inline it, so there won't be any difference at runtime.

src/main/java/blusunrize/immersiveengineering/api/shader/ShaderRegistry.java
```diff
@@ -29,12 +29,12 @@
 	/**
 	 * A map of shader name to ShaderRegistryEntry, which contains ShaderCases, rarity, weight and loot specifics
 	 */
-	public static LinkedHashMap<String, ShaderRegistryEntry> shaderRegistry = new LinkedHashMap<>();
+	public static LinkedHashMap<ResourceLocation, ShaderRegistryEntry> shaderRegistry = new LinkedHashMap<>();
```
malte0811: Changing all of these fields to the "basic" data structure (`Map<...>`, `List<...>`, etc) would probably be a good idea.

BluSunrize: Only on the left, I presume?

BluSunrize:
Could you detail "flattening" the shader item?
Would you recommend we take registration of entries out of it?

src/main/java/blusunrize/immersiveengineering/api/shader/ShaderRegistry.java
malte0811: Yes. How would you even change it on the RHS?!

malte0811:
Flattening=splitting an item with a small-ish number of states into one item per state. In this case one item per shader. It's recommended for 1.13+, but (as I said) incompatible with using a registry for shader types.

BluSunrize:
Hmm, I can't say I super like the idea of a single item per shader. Feels very...wasteful?

malte0811:
It's kind of the opposite of that. In 1.14 we essentially have `Integer.MAX_VALUE` items IDs available to us, so "wasting" a few 10 IDs shouldn't change much of anything. If we have one item per shader, a shader item stack is essentially a couple of `int`s and a `null` for the NBT. If we have one item for all shaders we still have the `int`s, but also a map of some sort for the NBT. One potential problem would be a slight increase in static memory usage (a copy of the model for each shader), but I doubt that will have any measurable effect (admittedly the savings from simpler shader item stacks most likely won't be measurable either).

BluSunrize:
It would make it more difficult for other mods to add shaders via the registry however, or add them from IE's compat module (like the Botania shader)
I do like the NBT solution for that.

malte0811:
It shouldn't be a problem for other mods (they can register shaders during mod construction), but to cover (nearly) all use cases cleanly we could add something like a `RegisterShaderEvent` that we would fire before adding the shader items, similar to what we would get for a Forge registry.

BluSunrize:
Eh, shitty rebase job, but it works and github is happy

Merging this now

PR #3625, opened 2019-12-30
Cleanup wire raytrace

malte0811:
Much cleaner code for the wire raytracer, as well as some basic unit tests for it.


BluSunrize:
Looks good to me. I do like the fact that this code identifies all colliding blocks and highlights them. That way, players don't have to keep trying and keep finding out what's in the way.

Changed approved.

PR #3676, opened 2020-03-22
Refactoring Shaders on 1.14

BluSunrize:
Made all shaders work again, refactored it all to be nicer, sleaker and match the obj callback structure.

It's still a nasty hack, but what can you do? ¯\_(ツ)_/¯


src/main/java/blusunrize/immersiveengineering/api/shader/ShaderLayer.java
```diff
@@ -0,0 +1,98 @@
...
+		this.color = new Vector4f((color >> 16&255)/255f, (color >> 8&255)/255f, (color&255)/255f, (color >> 24&255)/255f);
+		if(ShaderRegistry.defaultLayerBounds.containsKey(texture))
+			this.setTextureBounds(ShaderRegistry.defaultLayerBounds.get(texture));
+	}
+
+	public ShaderLayer setTextureBounds(double... bounds)
+	{
+		if(bounds==null)
+			return this;
+		assert (bounds.length==4);
```
malte0811: I prefer using `Preconditions.checkArgument` over `assert`s (at least for cheap things), since the former will always run and catch issues that might crash at any point early even in production (IIRC assert is disabled there by default).

PR #3697, opened 2020-03-25
Move Cloche to a Recipe System

BluSunrize:
My initial idea for the cloche was this cool handler based system for compat, and in the end it was just people complaining about lack of Craftweaker and JEI compat.

This is the better system.

**ToDo** (after this PR passes review and gets merged):
- Write Craftweaker compat
- Change any reference of "Belljar" to "Cloche" to clean up terminology


src/main/java/blusunrize/immersiveengineering/client/utils/ClocheRenderHelper.java
```diff
@@ -0,0 +1,217 @@
...
+			return 1;
+		}
+
+		@Override
+		public Collection<Pair<BlockState, TRSRTransformation>> getBlocks(ItemStack stack, float growth)
+		{
+			Set<Pair<BlockState, TRSRTransformation>> ret = new HashSet();
+			Matrix4f stemMatrix = new Matrix4f();
+			stemMatrix.setIdentity();
+			stemMatrix.rotY(-1.571f); //-90 deg
```
malte0811: Something like `-(float)Math.PI/2` would be clearer IMO.

src/main/java/blusunrize/immersiveengineering/client/utils/ClocheRenderHelper.java
```diff
@@ -0,0 +1,217 @@
...
+			stemMatrix.setTranslation(new Vector3f(.75f, .0625f, 0));
+
+			if(growth < .375)
+			{
+				int age = Math.min(7, Math.round(7*growth/.375f));
```
malte0811: `Math.min` shouldn't be necessary here, the second parameter is always <=7.

src/main/java/blusunrize/immersiveengineering/common/IERecipes.java
```diff
@@ -521,7 +565,7 @@ public static ArcFurnaceRecipe addArcRecipe(ItemStack output, Object input, int
 
 	public static void addArcOreSmelting(ItemStack output, String ore)
 	{
-		if(ApiUtils.isNonemptyItemTag(getOre(ore)))
+		if(ApiUtils.isNonemptyItemTag(getOre(ore)))//ToDo this should allow blocks too
```
malte0811: Why? Every block tag should have an equivalent item tag.

src/main/java/blusunrize/immersiveengineering/client/utils/ClocheRenderHelper.java
BluSunrize: I didn't want to include unncecessary maths, but I suppose Java optimises this to a constant anyway, huh?

src/main/java/blusunrize/immersiveengineering/common/IERecipes.java
BluSunrize: See my fix in src/main/java/blusunrize/immersiveengineering/api/ApiUtils.java#82

It doesn't seem to always be cohesive?

src/main/java/blusunrize/immersiveengineering/client/utils/ClocheRenderHelper.java
malte0811: I'm quite sure it actually optimizes it at compile-time (not JIT), ForgeFlower (Forge's FernFlower fork) "recently" had to add something to replace that sort of thing with something readable.

PR #3722, opened 2020-03-28
Adding Minecarts for basic IE blocks

BluSunrize:
Currently doing crates and barrels.

I'll consider capacitors in the future, but for that we'll need a dedicated loader block to plug them in


src/main/java/blusunrize/immersiveengineering/common/entities/IEMinecartEntity.java
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	public abstract void writeTileToItem(ItemStack itemStack);
+
+	public abstract void readTileFromItem(LivingEntity placer, ItemStack itemStack);
+
+	@Override
+	public Type getMinecartType()
+	{
+		return null;
```
SkySom: Pretty sure this is a Nonnull field. Just make it a Chest or Empty.

BluSunrize:
Sky approved it, ya'll saw it.

PR #3736, opened 2020-03-30
Capability reference invalidation proposal

stfwi:
Prevents capability references from inhibiting neighbour TEs to be properly removed by invalidating the references when a post placement update is invoked from the neighbour block.

Test setup: Barrel on the ground, above a valve, above another barrel. The valve block is broken while filling, and placed again (multiple times).

a) with IE .95, the log shows `fill()` calls in the neighbour fluid handler.
b) patched: the capability is checked again after valve block updates, resolving the issue.

[a-dangling-fluidhandler-tracelog.zip](https://github.com/BluSunrize/ImmersiveEngineering/files/4402036/a-dangling-fluidhandler-tracelog.zip)

[b-with-cap-ref-invalidation.zip](https://github.com/BluSunrize/ImmersiveEngineering/files/4402037/b-with-cap-ref-invalidation.zip)

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/IEBlockInterfaces.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/IEBlockInterfaces.java
@@ -449,4 +449,15 @@ public class IEBlockInterfaces
 		IModelData getModelData(@Nonnull IEnviromentBlockReader world, @Nonnull BlockPos pos, @Nonnull BlockState state,
 								@Nonnull IModelData tileData);
 	}
+
+	public interface ICapabilityReferenceHolder
+	{
+		void invalidateCapabilityReference(@Nullable Direction side);
+
+		default void invalidateCapabilityReferences()
+		{
+			invalidateCapabilityReference(null);
+		}
+	}
+
 }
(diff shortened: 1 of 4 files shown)


malte0811:
This should not be necessary if the other block uses `LazyOptional` correctly. Can you give me some steps for reproducing this issue?

stfwi:
- Steps as described in the test setup. Basically putting a wooden or metal barrel on top of something with a fluidhandler cap on top.
- The `LazyOptional` by the block used (here Pipe Valve where I added tracings) is returned in `getCapability()`. The TE does not `invalidate()` the `LazyOptional` in this case.

stfwi:
Let me add the cap invalidation and double check that.

stfwi:
Ok, Forge definitely invalidates caps, I thinks that we can take as a reference.
I suppose this PR can be closed -

stfwi:
For completeness, here the log where the neighbour TE invalidates caps on `remove()`. It works fine then.
[neighbour-lazy-optional-invalidation.zip](https://github.com/BluSunrize/ImmersiveEngineering/files/4402508/neighbour-lazy-optional-invalidation.zip)

Issue #3769, opened 2020-04-09
[0.14-95-c839561][request] Hemp recipe for Hammer?


iammeeper:
I searched through the issues - but as my search with only "hammer" without any other filter returnd 233 results I didn't looked through all of them - so if this is a duplicate it's not intential - just laziness not to search through 200+ issues.

TLDR: I started a new world with 0.14-95-c839561 and just got the idea: Hey, wasn't there a way to get String from Hemp or to craft the Hammer with Hemp directly? Well, guess this was some tweaks by some modpacks back in 1.12.2. So, I guess crafting String from Hemp is maybe too overkill how about a recipe about crafting the Hammer with Hemp in addition (or maybe instead of) to the current recipe with String?

iammeeper:
So, as I look through `Recipes.java` to come up with my own pull request I figured that Hemp Seeds no longer can be found by harvesting tall grass but only by craft a seed with a string, wich brings me back round to my initial issue: You have to somehow get a String to start IE, or two that is: one for the Hammer and one for the initial Hemp Seed. So, as we already talking about re-adding a way to craft String from Hemp Fiber and also the Hammer with Hemp Fiber how about also re-add the "getting Hemp Seed from harvesting tall grass"?
//also cross-posted in pull request #3771

gyroplast:
> You have to somehow get a String to start IE, or two that is: one for the Hammer and one for the initial Hemp Seed.

With the "3 hemp fiber -> 1 string" recipe added back in that was plain missing, we're back to needing to obtain exactly one string right now, to craft the hemp seed. This can be cultivated and turned into arbitrary amounts of additional string. Making a new recipe just for crafting the hammer with hemp fiber seems a bit misdirected, in my opinion.

I, too, hate having to fight spiders or spelunk through an abandoned mineshaft just to get this first piece of string, so having the seeds drop from tall grass as in the past is preferable. Commit a547abb5 hints at this being somehow troublesome with Forge currently, though.

BluSunrize:
Yeah we literally can't add drops to tall grass atm.
The string + seed recipe is a temporary solution to make hemp accessible to users until Forge fixes this.

iammeeper:
I agree with you that with the Hemp Fiber > String recipe merged my initial idea about adding a recipe for crafting the Hammer with Hemp Fiber is redundant. But I'm glad to see where it led to in the end.
[...]

PR #3771, opened 2020-04-10
[1.14.4] re-add torch and string recipes

gyroplast:
Like @iammeeper realized in issue #3769, the hemp fiber to string recipe got thrown under the bus with commit d3728ee7, as well as the recipe to craft torches from wool, wooden rods and creosote. I also really like those recipies as early/mid-game sinks for creosote and hemp fiber.

This PR re-adds both recipies, validated manual and assembler crafting.

### DIFF
--- a/src/generated/resources/data/immersiveengineering/advancements/recipes/decorations/torch.json
+++ b/src/generated/resources/data/immersiveengineering/advancements/recipes/decorations/torch.json
@@ -1,0 +1,54 @@
+{
+  "parent": "minecraft:recipes/root",
+  "rewards": {
+    "recipes": [
+      "immersiveengineering:torch"
+    ]
+  },
+  "criteria": {
+    "has_wool": {
+      "trigger": "minecraft:inventory_changed",
+      "conditions": {
+        "items": [
+          {
+            "tag": "minecraft:wool"
+          }
+        ]
+      }
+    },
+    "has_stick": {
+      "trigger": "minecraft:inventory_changed",
+      "conditions": {
+        "items": [
+          {
+            "tag": "forge:rods/wooden"
+          }
+        ]
+      }
+    },
+    "has_creosote": {
+      "trigger": "minecraft:inventory_changed",
+      "conditions": {
+        "items": [
+          {
+            "item": "immersiveengineering:creosote_bucket"
+          }
+        ]
+      }
+    },
+    "has_the_recipe": {
+      "trigger": "minecraft:recipe_unlocked",
+      "conditions": {
+        "recipe": "immersiveengineering:torch"
+      }
+    }
+  },
+  "requirements": [
+    [
+      "has_wool",
+      "has_stick",
+      "has_creosote",
+      "has_the_recipe"
+    ]
+  ]
+}
--- a/src/generated/resources/data/immersiveengineering/advancements/recipes/misc/string.json
+++ b/src/generated/resources/data/immersiveengineering/advancements/recipes/misc/string.json
@@ -1,0 +1,32 @@
+{
+  "parent": "minecraft:recipes/root",
+  "rewards": {
+    "recipes": [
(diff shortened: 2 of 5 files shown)


iammeeper:
@malte0811 Thank you for consider this change instead of rejecting it as too overpowered.

BluSunrize:
It's not overpowered at all, it's just missing, lol.

Make the changes to the Recipes class, and I'll merge this.

iammeeper:
So I guess this then has to go at blusunrize.immersiveengineering.common.data.Recipes.recipesMisc(Consumer) at line 1703?

BluSunrize:
No, if anything it should be in `recipesIngredients`, since that's where we put the recipe for Gunpowder. But it might actually be worthwhile adding a `recipeVanilla` method to put all that stuff in. To categorize it better.

iammeeper:
> No, if anything it should be in `recipesIngredients`, since that's where we put the recipe for Gunpowder. But it might actually be worthwhile adding a `recipeVanilla` method to put all that stuff in. To categorize it better.

I just had another look in that file for the Hammer recipe wich is currently at line 1135 in recipesTools. I'm not a developer, but I guess I may would be able to at least copy'n'paste it and change the F key to Hemp (I guess I'll find it somewhere else in that file to also just copy'n'paste), but neither would I be able to do this for that Torch recipe nor do I understand the second to last line about that addCriterion(). What I was able to find using Google it seems it has something to do with advancements and what is required for the crafting, but no full documentation of this function. Is it required for some specific reason?

Well, I think I don't have to bother anymore as this isn't my pull request. So hopefully @gyroplast does what's required to get this merged. Thanks for all the replies and not rejecting it.

iammeeper:
So, as I look through `Recipes.java` to come up with my own pull request I figured that Hemp Seeds no longer can be found by harvesting tall grass but only by craft a seed with a string, wich brings me back round to my initial issue: You have to somehow get a String to start IE, or two that is: one for the Hammer and one for the initial Hemp Seed. So, as we already talking about re-adding a way to craft String from Hemp Fiber and also the Hammer with Hemp Fiber how about also re-add the "getting Hemp Seed from harvesting tall grass"?
//also cross-posted in issue #3769

gyroplast:
Thank you @malte0811 for catching my obvious mistake here. I squashed/rebased the PR to look neater, and now it should re-introduce string and torch recipes

- without overwriting the vanilla recipe for torches
- working with the jerrycan (and likely all other fluid containers)
- with sensible vanilla recipe book support, albeit slightly incomplete

All in all this PR should do the job for now and give us back the missing recipes. I took up @BluSunrize suggestion of putting the new recipes in their own recipeVanilla() method as well.

A known, remaining issue with the vanilla recipe book is that choosing the torch recipe does _not_ pre-fill the crafting grid with a creosote bucket or jerrycan, even if available in the inventory. Placing either manually works fine, as does the assembler, and the exact same issue persists with the plain old treated wood planks. I'll add a new issue for that, however, to keep focus here.

Thanks for helping me get my bearings with MC modding, reliable Forge documentation is annoyingly scarce.

BluSunrize:
Looks pretty good to me, but I'd love it if you moved the recipe for gunpowder from `recipesIngredients` to `recipesVanilla`, for consistency ^_^

gyroplast:
> Looks pretty good to me, but I'd love it if you moved the recipe for gunpowder from `recipesIngredients` to `recipesVanilla`, for consistency ^_^

Sounds perfectly reasonable, happy to oblige. Rebased, pushed and tested, thanks!

BluSunrize:
I'm hereby dismissing Malte's review, because the point of critique has been resolved.

iammeeper:
Happy to see this simple request made it into official repo now, gonna clone and build current commit and start all over again to see if anything similar may comes to my attention. May I then open another issue?

PR #3773, opened 2020-04-10
patch retrogen, requires custom forge

phit:
depends on https://github.com/MinecraftForge/MinecraftForge/pull/6610

probably wouldnt merge this for now, but may be useful on 1.15..

I just really needed retrogen to work on my server on 1.14

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/world/IEWorldGen.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/world/IEWorldGen.java
@@ -22,7 +22,6 @@ import net.minecraft.world.biome.Biome;
 import net.minecraft.world.chunk.ChunkStatus;
 import net.minecraft.world.dimension.DimensionType;
 import net.minecraft.world.gen.GenerationStage.Decoration;
-import net.minecraft.world.gen.Heightmap;
 import net.minecraft.world.gen.feature.ConfiguredFeature;
 import net.minecraft.world.gen.feature.Feature;
 import net.minecraft.world.gen.feature.OreFeatureConfig;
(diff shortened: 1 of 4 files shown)


malte0811:
Thanks for the PR!

You'll need to fix your formatting (there's a formatting file for IntelliJ in the contributing guidelines). Other than that it's probably good.

If you want this in IE 1.14, feel free to add another coremod for it to IE, we already have one for https://github.com/MinecraftForge/MinecraftForge/pull/6519 and, seeing the progress on that PR, I doubt we'll get your Forge PR any time soon.

phit:
Sorry, no idea how core modding with ASM works, much less how it works with the new JS system.
Fixed the formatting! 
Yeah, Forge doesn't seem to care much about maintaining 1.14 even though they announced it would be LTS?

phit:
running this on our 1.14 now with a custom forge build, there's definitely still some odd stuff going on, need to move the writing of the tag that the chunk is done to after the generation
on the save event doesn't seem to be enough seeing lots of duplicate retrogen

malte0811:
What's the current state of this? I would like to have it in the next release. I can add the coremod myself if you say that you don't have any experience with them.

phit:
go ahead, I rebased it to latest

src/main/java/blusunrize/immersiveengineering/common/world/IEWorldGen.java
```diff
@@ -50,15 +53,44 @@ public static void addOreGen(String name, BlockState state, int maxVeinSize, int
...
+				for(Entry<String, ConfiguredFeature<?>> gen : retroFeatures.entrySet())
+				{
+					boolean retrogen = false;
+					try
+					{
+						retrogen = retrogenMap.get("retrogen_"+gen.getKey());
+					} catch(NullPointerException e)
+					{
+						retrogen = false;
+					}
```
malte0811, line 85: This block seems strange... Why are you catching NPEs rather than just checking whether something is `null`?

phit: no reason, just leftover from debugging, as `retrogenMap` was completely empty before this pr

Issue #3767, opened 2020-04-08
[1.14.4] Industrial Hemp Harvesting Noise


Hoskja:
Description:

I've noticed that hemp plants seem to make noises similar to stone breaking when broken. This is a repeated issue.

Debug Crash Report: 

https://pastebin.com/c2BpFc9m

Version/Mods:

minecraft version 1.14.4
mods: Reliquary version 1.3.4.947
IE version 0.14-94
forge: 1.14.4-28.2.3

PR #3775, opened 2020-04-10
emit crop sound on breaking planted hemp

gyroplast:
Change block break sound from default stone sound to crop. Addresses issue #3767.


BluSunrize:
Thanks! =)

PR #3780, opened 2020-04-10
Readding Chutes

BluSunrize:
What's scary and  goes *bonk* in the night?
That's right, chutes. They're scary because of how messy their code is.

That said, they're no longer conveyor subtypes, so that cleans things up a bunch.
They also do a bunch of checks now to avoid players getting stuck. It took me a while to get it all worked out, but now I'm at a stage where all my test setups didn't glitch me out anymore.


src/main/java/blusunrize/immersiveengineering/common/blocks/metal/ChuteTileEntity.java
```diff
@@ -0,0 +1,341 @@
...
+		{
+			BlockPos target = getPos().offset(facing);
+			TileEntity targetTile = world.getTileEntity(target);
+			boolean opposedChute = targetTile instanceof ChuteTileEntity&&((ChuteTileEntity)targetTile).diagonal;
+			// We're not going to redirect into another diagonal chute, to avoid infinite bouncing
+			if(opposedChute)
+				return;
+
+			long time = System.currentTimeMillis();
+			String hash = Integer.toHexString(getPos().hashCode());
```
malte0811, line 116: This seems strange. The only reason for using a hash (that I can think of) would be reduced memory usage in the NBT, but then turning it into a string will probably use more memory, not less. I'd suggest storing `BlockPos#toLong`, since that is unique for vanilla (not for CubicChunks, but still quite good).

src/main/java/blusunrize/immersiveengineering/common/data/BlockStates.java
```diff
@@ -375,6 +380,14 @@ protected void registerStatesAndModels()
 		createMultiblock(MetalDevices.teslaCoil, obj("block/metal_device/teslacoil.obj"),
 				null, IEProperties.MULTIBLOCKSLAVE, IEProperties.FACING_ALL, null,
 				180);
+		for(Entry<EnumMetals, Block> chute : MetalDevices.chutes.entrySet())
+		{
+			ModelFile model = ieObj("block/metal_device/chute_"+chute.getKey().tagName(), rl("block/metal_device/chute.obj.ie"))
+					.texture("texture", rl("block/metal/sheetmetal_"+chute.getKey().tagName()));
+//			createRotatedBlock(chute.getValue(), state -> model,
+//					IEProperties.FACING_HORIZONTAL, ImmutableList.of());
```
malte0811, line 388: Delete?

src/main/resources/assets/immersiveengineering/lang/en_us.json
```diff
@@ -824,6 +825,10 @@
   "block.immersiveengineering.conveyor_verticalcovered": "Vertical Covered Conveyor Belt",
   "block.immersiveengineering.conveyor_extract": "Extracting Conveyor Belt",
   "block.immersiveengineering.conveyor_extractcovered": "Covered Extracting Conveyor Belt",
+  "block.immersiveengineering.chute_iron": "Sheetmetal Chute",
+  "block.immersiveengineering.chute_copper": "Sheetmetal Chute",
+  "block.immersiveengineering.chute_aluminum": "Sheetmetal Chute",
+  "block.immersiveengineering.chute_steel": "Sheetmetal Chute",
```
malte0811, line 831: `Sheetmetal Chute (Steel)`?

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/ChuteTileEntity.java
```diff
@@ -0,0 +1,341 @@
...
+		if(this.diagonal&&entity.posY-getPos().getY() <= .625)
+		{
+			BlockPos target = getPos().offset(facing);
+			TileEntity targetTile = world.getTileEntity(target);
+			boolean opposedChute = targetTile instanceof ChuteTileEntity&&((ChuteTileEntity)targetTile).diagonal;
+			// We're not going to redirect into another diagonal chute, to avoid infinite bouncing
+			if(opposedChute)
+				return;
+
+			long time = System.currentTimeMillis();
```
malte0811, line 115: I'd recommend using the current tick instead, since that should have the same "logical" behavior even with lag/server restarts.

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/ChuteTileEntity.java
BluSunrize: I wanna say toLong wasn't around in 1.12, but I may be wrong. I'll change this though, for uniformity.

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/ChuteTileEntity.java
BluSunrize: Fair!

BluSunrize:
Re:

- Z-fighting on the item model: Do we have a way to only render selected parts on an item model? I didn't find an option for that, so I couldn't do anything about it.
- Breaking particles should be auto-determined by our code, but I'll check it over!
- Breaking textures are rendered on all faces for some reason, even hidden ones. You can see the same behavior on wooden posts. Appears to be a vanilla/forge issue.

malte0811:
> Do we have a way to only render selected parts on an item model?

I don't think so, at least not without making a custom item class. For some of the other items we have separate OBJs for block and item models, maybe that would be an option here as well?

BluSunrize:
Not a super fun solution, but for the sake of it looking nice, I'll do that.

BluSunrize:
Fixes implemented, awaiting second review

PR #3785, opened 2020-04-11
Make tall grass drop hemp seeds again

gyroplast:
Rather ugly but functional implementation to modify the loot table for tall grass to drop hemp seeds in addition to anything else it may drop, like vanilla wheat seeds.
Using one of the vanilla hoes increases drop chances by a factor of five.

Referring #3769, because @iammeeper and @BluSunrize may want to know it's not exactly impossible to make tall grass drop seeds again, but doing it _cleanly_ certainly requires a lot more effort and competence than I put into this initial "solution". Maybe it's a sufficient stop-gap measure for the next 1.14.4 release, though, or a starting point for someone more familiar with loot table shenanigans.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/EventHandler.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/EventHandler.java
@@ -58,11 +58,10 @@ import net.minecraft.util.math.Vec3d;
 import net.minecraft.util.text.StringTextComponent;
 import net.minecraft.world.World;
 import net.minecraft.world.dimension.DimensionType;
+import net.minecraft.world.storage.loot.LootPool;
+import net.minecraft.world.storage.loot.TableLootEntry;
 import net.minecraftforge.common.DimensionManager;
-import net.minecraftforge.event.AnvilUpdateEvent;
-import net.minecraftforge.event.AttachCapabilitiesEvent;
-import net.minecraftforge.event.TagsUpdatedEvent;
-import net.minecraftforge.event.TickEvent;
+import net.minecraftforge.event.*;
 import net.minecraftforge.event.TickEvent.WorldTickEvent;
 import net.minecraftforge.event.entity.EntityJoinWorldEvent;
 import net.minecraftforge.event.entity.living.*;
@@ -94,6 +93,8 @@ public class EventHandler
 	public static HashSet<IEExplosion> currentExplosions = new HashSet<IEExplosion>();
 	public static final Queue<Pair<DimensionType, BlockPos>> requestedBlockUpdates = new LinkedList<>();
 	public static final Set<TileEntity> REMOVE_FROM_TICKING = new HashSet<>();
+
+	private static final ResourceLocation TALL_GRASS_DROP = new ResourceLocation("minecraft", "blocks/tall_grass");
 
 	@SubscribeEvent
 	public void onLoad(WorldEvent.Load event)
@@ -176,6 +177,14 @@ public class EventHandler
 			new ResourceLocation(ImmersiveEngineering.MODID, "chests/village_blacksmith")
 	);
 	private static Field f_lootEntries;
+
+	@SubscribeEvent
+	public void lootTableLoad(LootTableLoadEvent event) {
+		if (event.getName().equals(TALL_GRASS_DROP))
+			event.getTable().addPool(LootPool.builder()
+					.addEntry(TableLootEntry.builder(new ResourceLocation(ImmersiveEngineering.MODID, "blocks/grass_drops")))
+					.name("ie_grass_drops").build());
+	}
 
 	/*TODO I think this can be done data-driven now?
 	@SubscribeEvent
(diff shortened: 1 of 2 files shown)


iammeeper:
Thanks for your work, may this gets merged or re-worked.

malte0811:
I agree with Blu, it seems very usable (and is apparently done by a few other mods already), but the hoe thing should be changed. I'm just a bit surprised that when I asked in the Forge discord whether this was possible yet, the answer was to wait for one of two PRs to be merged.

Additionally, I'd recommend using a more unique name for the loot pool, otherwise it will crash when another mod uses the same generic name.

TeamSpen210:
The problem is more that you can't modify anything in a consistent way. This doesn't entirely work, since it could potentially drop both normal seeds and hemp seeds.

BluSunrize:
So? Nothing wrong with dropping both...

gyroplast:
Thanks for your reviews! I got rid of the hoe logic entirely, and renamed the pool to `ie_grass_drops` to avoid name clashes. To credit, I totally got "inspired" by the [simplefarming mod](https://github.com/enemeez/simplefarming/blob/master/src/main/java/enemeez/simplefarming/events/LootTableHandler.java)'s implementation 
 of adding custom drops non-destructively to grass, fern, and tall grass, and thought if it works for them, why not for IE as well.

Again I validated that the hemp seed drops are _in addition to_ any other defined tall grass drops, and I figure that to be a sensible design decision and expected behaviour. I did not bother twiddling with the drop chance, weight, etc. and leave that to future balancing experts.
Right now, this JSON defines a 20% random condition for a uniform drop chance of 0-1 seeds, but this looting table seems to be applied twice upon breaking a 2-tall tall grass block, effectively leading to a rather attractive, non-uniform drop chance of P(0)=0.81, P(1)=0.18, and P(2)=0.01, if I didn't mix up something here. Looks fine to me, and felt reasonable when cutting down some plains in the game. :)

src/main/resources/data/immersiveengineering/loot_tables/blocks/grass_drops.json
```diff
@@ -0,0 +1,43 @@
+{
+	"type": "minecraft:block",
+	"pools": [
+		{
+			"name": "grass_drops",
```
BluSunrize, line 5: This pool-name does not match the name used in the event.

src/main/java/blusunrize/immersiveengineering/common/EventHandler.java
```diff
@@ -95,6 +94,8 @@
 	public static final Queue<Pair<DimensionType, BlockPos>> requestedBlockUpdates = new LinkedList<>();
 	public static final Set<TileEntity> REMOVE_FROM_TICKING = new HashSet<>();
 
+	private static final ResourceLocation tall_grass_drop = new ResourceLocation("minecraft", "blocks/tall_grass");
```
BluSunrize, line 97: Consider naming this constant in all-caps

gyroplast: Done.
Did not touch requestedBlockUpdates analogously, though, to prevent muddying the PR.

src/main/resources/data/immersiveengineering/loot_tables/blocks/grass_drops.json
gyroplast: True. Consistent now.

src/main/java/blusunrize/immersiveengineering/common/EventHandler.java
iammeeper: Naming what's called a constant (members with public, static and final (order of static and final can be interchanged)) in all uppercase spaced with underscore is suggested in the Java coding and naming conventions.

BluSunrize: ...yeah, I mean, that's kinda why I requested the change, lol.
You're a bit late there.

iammeeper:
from Forge changelog:

> Build: 1.14.4-28.2.4 - Sun Apr 12 20:56:37 GMT 2020
> 	David Quintana:
> 		Backport Global Loot Functions by Draco18s

So, I guess this may could be reworked?

iammeeper:
Addition (first wanted to open a new issue as I thought it would be somehow related to Forge 28.2.4 - but also occurs in 28.2.3): I noticed that "Tall Grass" in 1.14.4 seem to only refer to the 2-Block version instead of the "regular" 1-Block one. First I thought there's an issue with the backport to 28.2.4 but I only munched through 1-Block tall Tall Grass and didn't got anything. So I switched back to 28.2.3 and just by accident also chopped some 2-Block ones. As I noticed this I switched to 28.2.4 to confirm it's that and not the drop backport.
So: Is it intentional that "Tall Grass" only is the 2-Block version so only when you mow down the 2-Block version you even have a chance to get Hemp Seeds as the meaning of "Tall Grass" changed since 1.12.2? All that new things in 1.14.4 I'm not yet used to, guess that's one of them.

BluSunrize:
> 
> 
> from Forge changelog:
> 
> > Build: 1.14.4-28.2.4 - Sun Apr 12 20:56:37 GMT 2020
> > David Quintana:
> > Backport Global Loot Functions by Draco18s
> 
> So, I guess this may could be reworked?

Sure, feel free to rework it =p

PR #3802, opened 2020-04-13
Add the Item Batcher

BluSunrize:
It batches items based on its filter
Features:
- single & combined batching
- improved ghost slots
- improved state buttons for IE GUIs

This will be followed by another branch that implements the previously discussed additions to the coloured redstone system


src/main/java/blusunrize/immersiveengineering/client/gui/elements/GuiButtonState.java
```diff
@@ -14,21 +14,52 @@
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.FontRenderer;
 
-public class GuiButtonState extends GuiButtonIE
+public class GuiButtonState<E extends Enum> extends GuiButtonIE
```
malte0811, line 17: Are you actually using the `Enum` type bound for anything? If not you could just remove it as well as the special casing for `boolean`s.

BluSunrize: Uh, yeah, `GuiButtonDyeColor` uses it, and so does the button for the batching state in in `ItemBatcherScreen`
The whole point was to make creating those kinds of buttons easier ^^

malte0811: That's not what I mean. Is there any code that would break if I were to change `GuiButtonState<E extends Enum>` by `GuiButtonState<E>` and then create an instance where `E` is e.g. `String`? I haven't seen any, but I may have missed it.

BluSunrize: Hmm, no, I don't think there is?

BluSunrize: You're thinking that would be be a better way to handle the booleans, and also allow other things to by "cycled" in the future, without relying on enums?

malte0811: Yes.

BluSunrize:
How's this, Malte?

PR #3804, opened 2020-04-14
Bundled Redstone Capabilities

BluSunrize:
This adds the necessary API parts for blocks to interact with bundled redstone by way of the Redstone Interface Connector (or `ConnectorBundled`, internally)

Currently only the Item Batcher uses this, but other stuff could benefit from it in the future :D


BluSunrize:
ToDo: Manual Entry

src/main/java/blusunrize/immersiveengineering/common/blocks/wooden/ItemBatcherTileEntity.java
```diff
@@ -113,6 +119,19 @@ protected boolean isFilterMatched(int slot)
 				&&this.inventory.get(slot+9).getCount() >= this.inventory.get(slot).getCount();
 	}
 
+	protected Map<DyeColor, Boolean> calculateRedstoneOutputs()
+	{
+		Map<DyeColor, Boolean> map = new EnumMap<>(DyeColor.class);
```
malte0811, line 124: Why are you using a map to `boolean` rather than a `Set`/`EnumSet`?

BluSunrize: It was originally map to Integer, but yeah, Set is probably more sensible.

BluSunrize: Actually, no, a set is not sensible. This is an `AND` logic. With a set, you can only add or remove. If you have a filter not satisfied, but a later one of the same colour is, it'll add the dye to the set, despite not all filters being satisfied.

We can use `keySet().stream().filter()` to make it *return* a set of those with true for a value, but the initial calculation needs to use a map.

PR #3838, opened 2020-04-29
Simplify multiblock bounding box code

malte0811:
I've wanted to do this for some time now, finally got around to it. The main reason the multiblock bounding box code was so unreadable was because it did the transformations (mirror, rotation) explicitly for each box, rather than using some central code for it (and generating the boxes for a fixed orientation).  
The boxes are now generated for a non-mirrored multiblock facing north, and transformed to match the actual multiblock afterwards.

I don't expect a review of the per-multiblock changes (most of that was automatic refactoring by IDEA with some Search&Replace thrown in for `flipFront`/`flipRight`) but maybe of the "central" transformation code.


src/main/java/blusunrize/immersiveengineering/common/blocks/metal/MixerTileEntity.java
```diff
@@ -172,17 +170,14 @@ public void tick()
...
-							{
-								int drained = output.fill(Utils.copyFluidStackWithAmount(out, Math.min(out.getAmount(), accepted), false),
-										FluidAction.EXECUTE);
-								this.tank.drain(drained, FluidAction.EXECUTE);
-								ret = true;
-							}
+							int drained = output.fill(Utils.copyFluidStackWithAmount(out, Math.min(out.getAmount(), accepted), false),
+									FluidAction.EXECUTE);
+							this.tank.drain(drained, FluidAction.EXECUTE);
+							ret = true;
```
BluSunrize, line 180: It's a bit awkward to have this in the PR, but filtering it out now would be a pain too, so I guess we'll keep it

malte0811: That's what I get for telling IntelliJ to resolve all "constant boolean expression" problems in a file...

BluSunrize: And that's why I don't ever let IntelliJ auto-fix anything except imports. Because I had "rearrange entries" on once, and that was a mess.

malte0811: How else would you have dealt with simplifying all of that bounding box code? My approach was to make `fl`/`fw` final and set it to `Direction.NORTH`/`EAST`, and then let IntelliJ clean up all the expressions that had become constant. I don't think there was an option to do that just for one method.

PR #3846, opened 2020-05-05
Turntable Update

### LINKED ISSUE #3003: [Suggestion] Allow the Turntables redstone accepting sides to be inverted with the hammer
Sometimes you want to just switch between 2 orientations for i.e. conveyors, or avoid a certain direction when spinning it around. Since the turntable already has 4 sides with a "redstone dot", I suggest the ability to invert those, similar to the one of the big machines. Instead of changing it from active low to active high, the turntable would just spin in the opposite direction when receiving a redstone signal at this side.

MalkContent:
made turntable redstone input sided and enabled different turn direction depending on side receiving signal
added 180° turns
adjusted form and speed of visualization of turn direction to better mark the difference between 90° and 180° turns
prevented turntable blocks from forwarding redstone signals through it like a regular block

closes #3003


BluSunrize:
Please throw the logs you commited into the gitignore to remove them.

Also, do you feel comfortable adjusting the manual entry to explain the new behavior?

PR #3859, opened 2020-05-14
Adds Dropping Conveyor Redstone Control

MalkContent:
Prevents Dropping Conveyors from dropping/downwards inserting if they are receiving a redstone signal (regular conveyor insertions still apply)

small typo fix


BluSunrize:
LGTM

Issue #3854, opened 2020-05-12
[1.12.2] Crusher ticking block entity | arithmetic exception


DAmNRelentless:
### Description of the issue:
When you build a crusher, connect it to power and throw anything in, it will crash the game.
We currently use Immersive Engineering in our modpack "All The Mods 3 Expert Shedding Tiers" and multiple people reported this behaviour now. I've seen multiple issues are still open about this topic but none of them solves our problem. It was mentioned that it can be caused by multipliers in the config but they're all default for us and I even tried to delete the config so it is recreated but the crash still happens.

### Crashlog:
https://pastebin.com/UFnsJuF4

### Versions & Modlist
Game: 1.12.2
Forge: 14.23.5.2847
Immersive Engineering: 0.12-92

### Additional context
Our modpack is open source and can be found here in case you want to checkout our configuration: https://github.com/AllTheMods/ATM3-Expert
Everything that is script related to Immersive Engineering can be found here: https://github.com/AllTheMods/ATM3-Expert/blob/dev/overrides/scripts/mods/immersiveengineering.zs

DAmNRelentless:
Same issue as here: https://github.com/BluSunrize/ImmersiveEngineering/issues/3826

Issue #3826, opened 2020-04-21
[0.12-92] Crash - Ticking block entity - Crusher


jonathanbrooke:
**### Description of the issue:**
- Whenever an ore is tossed into the Crusher, the server immediately crashes.
- Cobblestone is accepted and works.

**### Crashlog:**
https://gist.github.com/jonathanbrooke/fe7ff84f6a9cd765952e7d3d8e1153be

**### Versions & Modlist**
IE Version = ImmersiveEngineering-0.12-92.jar   
Forge Version = 14.23.5.2847 

**###Other**
Mod is running on a Server & Client using Technic Launcher Deployment:
- Minecraft Version: 1.12.2
- Operating System: Linux (amd64) version 4.19-ovh-xxxx-std-ipv6-64
- Java Version: 1.8.0_211, Oracle Corporation
- Java VM Version: Java HotSpot(TM) 64-Bit Server VM (mixed mode), Oracle Corporation
- Memory: 3295136736 bytes (3142 MB) / 8239710208 bytes (7858 MB) up to 8589934592 bytes (8192 MB)

Happens every time conditions above are met

Let me know if you need more information.

Thanks!

BluSunrize:
Please provide your config file for Immersive Engineering.

jonathanbrooke:
IE Config Server - https://gist.github.com/jonathanbrooke/432a3145efc554598c31fa65a68b5a63
IE Config Client - https://gist.github.com/jonathanbrooke/4fba082369e4d7af567d875ba0a33ebc

jonathanbrooke:
To clarify, user on my server said this error is with the Crusher not the Pulverizer. I will edit accordingly, apologies!

BluSunrize:
Do you use Craftweaker to load custom recipes?

BluSunrize:
Nvm, you don't, I should have checked your modlist

BluSunrize:
Honestly, absolutely no idea what's happening there.
Can the player reproduce this issue in singleplayer?

jonathanbrooke:
@BluSunrize I've contacted the player who reported it to recreate in Singleplayer for you, ill revert when I've had confirmation he has done this.
Are there any further logs I can provide to assist?

jonathanbrooke:
@BluSunrize 
Single Player Crash Video from the user who reported it - https://youtu.be/hnZ5tpUJuAE
[...]

Issue #3417, opened 2019-03-16
[1.12.2] [0.12 - 89] Crusher causes server crash when ores inserted


eBuilder101:
After dropping ores into the crusher, the game suddenly crashed. I play singleplayer, and this was a server crash. The crash report mentions a ticking tile/block entity, which corrupted my world. I was able to use MCedit to remove the crusher. I tried using different insertion methods and different ores, but all resulted in this crash.

Report: https://hastebin.com/dobabekuqi.coffeescript

mindforger:
Looks like a broken recipe which causes a division by zero

eBuilder101:
What can be done about it?

BluSunrize:
Could you please show your IE config file?

eBuilder101:
Sure thing:

https://hastebin.com/ulixuceyut.coffeescript

It's modified a bit.

BluSunrize:
Yeah and that's what's causing it. You set the energy modifier for the crusher too low.

eBuilder101:
I tried with the default and it still crashed.

BluSunrize:
If you set `crusher_energyModifier` and `crusher_timeModifier` back to 1, you should see this bug fixed.

PR #3871, opened 2020-05-19
Update TileEntityMultiblockMetal.java

Thonikum:
This should fix the bug where settting the totalProcessEnergy for a custom crusher recipe to low(below 25) or messing with the crusher_timeModifier in the config can result in a division by zero during entity ticks. 

This should fix the following issues:
#3417
#3826
#3854
ICannt/Netherending-Ores#24

I fixed it by just checking for zero beforhand and then preventing the division.
This would cause the crushing to not consume any energy though, so feel free to use another workaround. 

This fix could also be applied to the 1.14 version in the file:
ImmersiveEngineering/src/main/java/blusunrize/immersiveengineering/common/blocks/generic/PoweredMultiblockTileEntity.java
in line 467


BluSunrize:
Thanks

BluSunrize:
For future reference, you should probably use a development environment the next time you PR anything, because you code didn't even compile =P

PR #3874, opened 2020-05-24
update dropping conveyor manual entry to include new redstone behavior

MalkContent:


MalkContent:
I did make sure that the text stays within the boundaries of the same page in the manual

PR #3938, opened 2020-06-09
Add shader pack compatibility to IEOBJItemRenderer.

Vaelzan:


src/main/java/blusunrize/immersiveengineering/client/render/IEOBJItemRenderer.java
```diff
@@ -124,38 +124,39 @@ else if(quadsForLayer.layer.isTranslucent())
...
 			IntBuffer intbuffer = bytebuffer.asIntBuffer();
 
-			for(int k = 0; k < j; ++k)
+			for(int i = 0; i < j; ++i)
 			{
 				intbuffer.clear();
-				intbuffer.put(aint, k*8, 8);
+				intbuffer.put(vertexData, i*intSize, intSize);
 				//TODO general formats?
 				float x = bytebuffer.getFloat(0);
```
malte0811, line 156: It might be a good idea to not rely on the order of the elements in the vertex format. We already have some [helper functions](https://github.com/BluSunrize/ImmersiveEngineering/blob/3226ec16e76dadbe02501d09e6908530897f0458/src/main/java/blusunrize/immersiveengineering/client/ClientUtils.java#L1049-L1057) to get the offsets for position and UV from the format.

src/main/java/blusunrize/immersiveengineering/client/render/IEOBJItemRenderer.java
```diff
@@ -124,38 +124,39 @@ else if(quadsForLayer.layer.isTranslucent())
...
+		int[] vertexData = quadIn.getVertexData();
 		Vec3i normalInt = quadIn.getFace().getDirectionVec();
 		Vector3f normal = new Vector3f(normalInt.getX(), normalInt.getY(), normalInt.getZ());
 		Matrix4f transform = matrixEntryIn.getMatrix();
 		normal.transform(matrixEntryIn.getNormal());
-		int i = 8;
-		int j = aint.length/8;
+
+		int intSize = DefaultVertexFormats.BLOCK.getIntegerSize();
+		int j = vertexData.length/intSize;
```
malte0811: As you already said on Discord, this is not an ideal name. I would suggest something like `numVertices`.

malte0811 (changes requested):
Looks good in general (and like it should work), I'm looking into the failed GH action now.

src/main/java/blusunrize/immersiveengineering/client/render/IEOBJItemRenderer.java
Vaelzan: In this case I was just trying to resolve this particular issue without making any major changes. As such, I left most of it the way I found it initially other than what was required to fix it (and a couple of minor tweaks). If I work on the code more frequently in the future I might make changes like that.

src/main/java/blusunrize/immersiveengineering/client/render/IEOBJItemRenderer.java
Vaelzan: That would work better, yes. I just named it somewhat after the method being called, which is also pretty poorly named really. I'll adjust that.

malte0811 (approved):

BluSunrize (approved):
LGTM

PR #3962, opened 2020-06-12
Configure Redstone Connectors with the Screwdriver

### LINKED ISSUE #3916: [Suggestion] Select Redstone Wire frequency/channel/colour differently
There's too many colours and it isn't possible to rotate them in the opposite direction. It's too easy to skip the desired colour by mistake and there's no way to go back, so I suggest maybe use the mousewheel to select them and/or use the screwdriver instead of the hammer, so that there aren't too many actions done by the hammer alone.

BluSunrize:
Using the Screwdriver instead of Hammer to configure redstone connectors. This now uses a small GUI to make the process quicker.

This closes #3916


malte0811 (changes requested):
The code looks good, I have one minor issue with the UI: Clicking on the same color twice will turn off all buttons. That is not a valid state for the connector (and exiting the GUI will result in a connector with the color you just turned off), so it shouldn't be possible to get the GUI into that state either.

BluSunrize:
oh shit, that's fair, yeah.
Will fix

BluSunrize:
Fixed!

malte0811 (approved):

PR #4009, opened 2020-06-19
Russian translation update for 1.15(blocks and items).

Sergo467:


BluSunrize:
Urm.
I don't changing the path to the translation file is a good idea. It needs to be in `resources/assets/[mod]/lang`
Your path uses cyrillic (?) symbols and has spaces in it. I don't think that'll work.

Sergo467:
Automatic translation of the page is messed up. The path must be src/main/resources/assets/immersiveengineering/lang/ru_ru.json

PR #4033, opened 2020-06-28
Move most methods in ApiUtils to smaller, more specific util classes

malte0811:
A few methods are left in `ApiUtils`, I didn't really know where else to put them. I've left the methods in their old order for now for simpler diffs, for the final merge it would probably be a good idea to move all the non-deprecated methods to the top.


BluSunrize (approved):
One compilation error in FluidPipes as a result of removing `Utils.toCC`, while I was adding a way to pull with pipes.
Want me to fix it?

The rest looks good to me.

BluSunrize:
Approved!

PR #4037, opened 2020-06-30
Fix a small bug with LittleTiles

FreezePhoenix:
This PR simplifies the models of the Coils, which allows better compatibility with some other mods.


BluSunrize:
1.12 is no longer receiving any updates

PR #4052, opened 2020-07-04
New Mineral System

BluSunrize:
New mineral vein system.

Still WIP, not ready for merge.

**Done:**
- Positioned veins with radius
- Vein generation with worldgen
- Configuration of noise threshold, total yield and initial yield reduction
- Distance based fail chance
- Overlapping veins (Results & Core samples)
- `mineral list` command
- `mineral get` command
- `mineral put` command
- `mineral setDepletion` command
- Maps sold by villagers

**Todo:**
- remove debug view (it causes CMEs and only works in singleplayer anyway)
- Survey tools item that makes finding veins easier


src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * @author BluSunrize - 03.06.2015
+ * <p>
+ * The Handler for the Excavator. Generates veins upon world gen, stores them, and keeps a cache of queried positions
+ */
+public class ExcavatorHandler
+{
+	private static final ArrayListMultimap<DimensionType, MineralVein> MINERAL_VEIN_LIST = ArrayListMultimap.create();
+	private static final HashMap<ColumnPos, MineralWorldInfo> MINERAL_INFO_CACHE = new HashMap<>();
```
malte0811: I don't remember whether we've talked about this before, but I prefer to use the most "abstract" type I can get away with for variables (within reason, so not `Collection<T>` over `Set<T>` in most cases). So this should be a `Multimap` and a `Map`, not an `ArrayListMultimap` and a `HashMap`.

Edit: Same for the accessor methods. Even more relevant there, in case we decide to switch to a slightly different type later on.

src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
```diff
@@ -0,0 +1,199 @@
...
+		int zStart = chunkpos.getZStart();
+		double d0 = 0.0625D;
+		boolean containsVein = false;
+		ColumnPos pos = null;
+		iteratechunk:
+		for(int xx = 0; xx < 16; ++xx)
+			for(int zz = 0; zz < 16; ++zz)
+			{
+				double noise = noiseGenerator.noiseAt((xStart+xx)*d0, (zStart+zz)*d0, d0, xx*d0);
+				double chance = Math.abs(noise)/.55;
```
malte0811, line 110: Where does this `/0.55` come from? And why isn't it part of the `mineralNoiseThreshold`?

src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
```diff
@@ -0,0 +1,199 @@
...
+				if(chance > mineralNoiseThreshold)
+				{
+					containsVein = true;
+					pos = new ColumnPos(xStart+xx, zStart+zz);
+					break iteratechunk;
+				}
+			}
```
malte0811, line 116: Slight complaint about the algorithm: Currently the probability of a block being a vein center decreases with increasing (chunk-relative) `x` and `z`. Would iterating the whole chunk and using the maximum value cause significant runtime issues?

src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
```diff
@@ -0,0 +1,199 @@
...
+		if(containsVein)
+		{
+			ColumnPos finalPos = pos;
+			int radius = 12+rand.nextInt(32);
+			int radiusSq = radius*radius;
+			boolean crossover = MINERAL_VEIN_LIST.get(dimension).stream().anyMatch(vein -> {
+				int dx = vein.getPos().x-finalPos.x;
+				int dz = vein.getPos().z-finalPos.z;
+				return dx*dx+dz*dz < vein.getRadius()*vein.getRadius()||dx*dx+dz*dz < radiusSq;
```
malte0811: `int distSq = dx*dx+dz*dz`? To avoid computing it twice, though I assume the JIT would be able to do that just as well.

src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
```diff
@@ -0,0 +1,199 @@
...
+			}
+		}
+	}
+
+	public static class MineralSelection
+	{
+		private final int totalWeight;
+		private final Set<MineralMix> validMinerals;
+
+		public MineralSelection(DimensionType dimension, int radius)
```
malte0811: What is the `radius`-parameter needed for?

src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
```diff
@@ -0,0 +1,199 @@
...
+			Set<MineralMix> surrounding = new HashSet<>();
+//			for(int xx = -radius; xx <= radius; xx++)
+//				for(int zz = -radius; zz <= radius; zz++)
+//					if(xx!=0||zz!=0)
+//					{
+//						DimensionChunkCoords offset = chunkCoords.withOffset(xx, zz);
+//						MineralWorldInfo worldInfo = mineralCache.get(offset);
+//						if(worldInfo!=null&&worldInfo.mineral!=null)
+//							surrounding.add(worldInfo.mineral);
+//					}
```
malte0811: Remove commented-out code.

src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
```diff
@@ -0,0 +1,199 @@
...
+		}
+
+		public int getTotalWeight()
+		{
+			return this.totalWeight;
+		}
+
+		public int getRandomWeight(Random random)
+		{
+			return Math.abs(random.nextInt()%this.totalWeight);
```
malte0811: `random.nextInt(this.totalWeight)`?

src/main/java/blusunrize/immersiveengineering/api/excavator/MineralVein.java
```diff
@@ -0,0 +1,107 @@
...
+	{
+		return mineral;
+	}
+
+	public int getRadius()
+	{
+		return radius;
+	}
+
+	// exponential growth of the fail chance depending on the distance from the center of the vein
```
malte0811: That still isn't exponential, and it never was. Maybe just remove the comment?

src/main/java/blusunrize/immersiveengineering/api/excavator/MineralVein.java
```diff
@@ -0,0 +1,107 @@
...
+
+	public void deplete()
+	{
+		depletion++;
+		IESaveData.setDirty();
+	}
+
+	public boolean isDepleted()
+	{
+		return ExcavatorHandler.mineralVeinYield==0||getDepletion() >= ExcavatorHandler.mineralVeinYield;
```
malte0811: Can depletion ever become negative? If it can't the first case is redundant.

src/main/java/blusunrize/immersiveengineering/client/ClientEventHandler.java
```diff
@@ -1282,6 +1281,59 @@ public void onRenderWorldLastEvent(RenderWorldLastEvent event)
...
+				tessellator.draw();
+
+				BufferBuilder.setTranslation(0, 0, 0);
+			}
+			GlStateManager.shadeModel(GL11.GL_FLAT);
+			GlStateManager.enableCull();
+			GlStateManager.disableBlend();
+			GlStateManager.enableTexture();
+		}
+
```
malte0811, line 1327: I assume this is the debug code that needs to be remove/disabled?

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/SampleDrillTileEntity.java
```diff
@@ -306,6 +289,7 @@ else if(!this.active)
 		return false;
 	}
 
+	/* TODO: I think these were for computer integration?
```
malte0811: Either that or for something else that was removed/became obsolete at some point. Feel free to remove them.

src/main/java/blusunrize/immersiveengineering/common/items/CoresampleItem.java
```diff
@@ -52,57 +63,76 @@ public CoresampleItem()
...
-			MineralMix mineral = getMix(stack);
-			if(mineral!=null)
-			{
-				String unloc = mineral.getTranslationKey();
-				String loc = I18n.format(unloc);
-				if(unloc.equals(loc))
-					loc = mineral.getPlainName();
-				list.add(new TranslationTextComponent(Lib.CHAT_INFO+"coresample.mineral", loc));
-			}
+			ListNBT nbtList = coresample.getOrCreateTag().getList("mineralInfo", NBT.TAG_COMPOUND);
```
malte0811, line 77: Why isn't this using `getMineralMixes`?

src/main/java/blusunrize/immersiveengineering/common/items/SurveyToolsItem.java
```diff
@@ -0,0 +1,230 @@
...
+		});
+		if(tooClose)
+		{
+			player.sendStatusMessage(new TranslationTextComponent(Lib.CHAT_INFO+"survey.too_close"), true);
+			return stack;
+		}
+
+		// Get angle between postion->center vector and standard (south facing) vector
+		Vector2f vecToCenter = new Vector2f(vein.getPos().x-pos.getX(), vein.getPos().z-pos.getZ());
+		double angle = Math.toDegrees(Math.atan2(vecToCenter.y, vecToCenter.x));
```
malte0811: What does this do when you sample at the vein center?

malte0811 (changes requested):

src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
BluSunrize: Vanilla's Perlin Noise generator multiplies its result with 0.55. This was to counter it =P

src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
BluSunrize: Probably not a dreadful loss in performance, and you have a good point. I'll see to it.

src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
BluSunrize: Fair point. Even if Java fixes it internally, we can be explicit here.

src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
BluSunrize: Nothing anymore! Legacy stuff I forgot to remove :D

src/main/java/blusunrize/immersiveengineering/api/excavator/MineralVein.java
BluSunrize: I'm bad at maths! D:

src/main/java/blusunrize/immersiveengineering/api/excavator/MineralVein.java
BluSunrize: No, this is actually a fault in my logic. If mineralVeinYield is 0, all veins should be infinite. Will fix.

src/main/java/blusunrize/immersiveengineering/client/ClientEventHandler.java
BluSunrize: Yup.

src/main/java/blusunrize/immersiveengineering/api/excavator/MineralVein.java
BluSunrize: Wait, correction. If the yield is -1, then it's infinite. Says so in the config.

src/main/java/blusunrize/immersiveengineering/common/items/CoresampleItem.java
BluSunrize: Because I need the other values in the NBT tag. `getMineralMixes` returns only the mixes, but I need saturation and depletion here.

src/main/java/blusunrize/immersiveengineering/common/items/SurveyToolsItem.java
BluSunrize: It puts out "0 steps to the east". Could probably specialcase it.

malte0811: Would probably be a good idea from a UI perspective. I was a more worried about having `NaN`s/`Infinity`s running around there, but it looks like `atan2` just returns 0 for `(0, 0)`.

src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
malte0811: Ok, then I'm fine with keeping the factor, but I think it should be documented somewhere.

src/main/java/blusunrize/immersiveengineering/api/excavator/ExcavatorHandler.java
```diff
@@ -188,7 +180,7 @@ public int getTotalWeight()
 
 		public int getRandomWeight(Random random)
 		{
-			return Math.abs(random.nextInt()%this.totalWeight);
+			return Math.abs(random.nextInt(this.totalWeight));
```
malte0811: `abs` is no longer necessary here.

malte0811 (changes requested):

malte0811 (approved):

PR #4097, opened 2020-07-17
Blocking conveyors

malte0811:
Make conveyor belts not accept more items when 3 items/entities are on the belt already, or have been in the last 10 ticks. The second part is needed to work around the fact that item entities only call the collision method every 4th tick, so we need to remember at least 4 ticks, and I increased that to 10 just to be safe.


src/main/java/blusunrize/immersiveengineering/api/tool/ConveyorHandler.java
```diff
@@ -67,6 +72,36 @@
 	public static final Map<ResourceLocation, Block> conveyorBlocks = MetalDevices.CONVEYORS;
 	public static final ResourceLocation textureConveyorColour = new ResourceLocation("immersiveengineering:block/conveyor/colour");
 
+	// Should work for multiple dimensions since the calls aren't "interleaved" for multiple dimensions
+	private static final IntSet entitiesHandledInCurrentTickClient = new IntOpenHashSet();
+	private static final MutableLong currentTickClient = new MutableLong();
+	private static final IntSet entitiesHandledInCurrentTickServer = new IntOpenHashSet();
+	private static final MutableLong currentTickServer = new MutableLong();
+
+	// Returns true iff the entity has not been handled before in this tick
```
BluSunrize, line 81: Your if has a spare f =P

src/main/java/blusunrize/immersiveengineering/api/tool/ConveyorHandler.java
```diff
@@ -363,57 +398,101 @@ else if(getConveyorDirection()==ConveyorDirection.DOWN)
...
-			if(!isActive())
+			if(!isActive()||entity==null)
 				return;
+			boolean hasBeenHandled = !markEntityAsHandled(entity);
 			BlockPos pos = getTile().getPos();
 			ConveyorDirection conveyorDirection = getConveyorDirection();
 			float heightLimit = conveyorDirection==ConveyorDirection.HORIZONTAL?.25f: 1f;
-			if(entity!=null&&entity.isAlive()&&!(entity instanceof PlayerEntity&&entity.isSneaking())&&entity.posY-pos.getY() >= 0&&entity.posY-pos.getY() < heightLimit)
+			final boolean outputBlocked = isOutputBlocked();
+			if(entity.isAlive()&&!(entity instanceof PlayerEntity&&entity.isSneaking())&&entity.posY-pos.getY() >= 0&&entity.posY-pos.getY() < heightLimit)
```
BluSunrize: This if condition should be broken up for readability. Checks for "isAlive" and the player check can be changed to return earlier.

BluSunrize (approved):
Minor remarks, the rest looks good to me.

Tests done in SSP were sucesful for me

src/main/java/blusunrize/immersiveengineering/api/tool/ConveyorHandler.java
malte0811: No, it doesn't:
 > [if and only if (shortened as iff)](https://en.wikipedia.org/wiki/If_and_only_if)

src/main/java/blusunrize/immersiveengineering/api/tool/ConveyorHandler.java
malte0811: I mean, it's already an improvement over what was there before. The only reason I moved the null-check is that I access the entity before that, so it can't ever be null at that point. But yes, I'll change it.

PR #4107, opened 2020-07-19
Concept for Version Numbering

BluSunrize:
**This is not getting merged**

I just wanted to give @malte0811 an idea for how it looks =p


BluSunrize:
**Concept for "Milestones", and resulting version numbers:**

I think it's valid for us to ignore any changes before 1.14.4. That update changed a lot of internal structures, class organization and made large changes to the API as well. So we could label is as "1.0" of the IE API as it stands now.

We had another major change in 0.14-96, when we migrated everything to JSON recipes, which most definitely broke the API. So that would put us at Version 2.0

Version 0.14-104 (and its parallel builds 105 and 106) are the ones that changed the Mineral system completely, raising our API version to 3.0.

Since then I don't think we've had breaking changes, since your conveyor PR was written to make correct use of deprecation. So we can treat 107/108/109 as content updates.

That would currently put us at Version **3.1** (or 3.2 on 1.16, if we are considering integration of Netherite as a content update).
Opinions?

PR #4137, opened 2020-07-27
Fix a crash if wall Shader Banners are placed directly without NBT, and fix collisions

TeamSpen210:
Shader banners have a `wall` boolean, which identifies which type of banner it is. Now that 1.13+ use two different blocks, this means that if it doesn't match the block it will crash in the renderer. Instead just check the blockstate directly.

The issue can be easily replicated by either doing `/setblock ~ ~ ~ immersiveengineering:shader_banner_wall` or creating a Debug world and flying over to bring it into render range.

I did also notice that the collision shape is inconsistent with vanilla, and so fixed that too. I moved the method to the blocks since the TE no longer affects this. Separate commit if you'd prefer to just cherry-pick the first error

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/client/render/tile/ShaderBannerRenderer.java
+++ b/src/main/java/blusunrize/immersiveengineering/client/render/tile/ShaderBannerRenderer.java
@@ -38,8 +38,11 @@ public class ShaderBannerRenderer extends TileEntityRenderer<ShaderBannerTileEnt
 	{
 		long time = te.getWorldNonnull().getGameTime();
 		GlStateManager.pushMatrix();
+
+		// Check which of the two blocks we are so we can calculate the orientation.
 		if(te.getState().getBlock() == Cloth.shaderBanner)
 		{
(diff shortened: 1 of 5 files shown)


src/main/java/blusunrize/immersiveengineering/client/render/tile/ShaderBannerRenderer.java
```diff
@@ -37,7 +38,7 @@ public void render(ShaderBannerTileEntity te, double x, double y, double z, floa
 	{
 		long time = te.getWorldNonnull().getGameTime();
 		GlStateManager.pushMatrix();
-		if(!te.wall)
+		if(te.getState().getBlock() == Cloth.shaderBanner)
```
BluSunrize, line 43: Throw a comment here to make it clear what that condition does, please.
`wall` was self-explanatory, this one is a little more obscure.

BluSunrize (approved):
Looks good to me, except for one minor remark =)

BluSunrize:
Looks good!

PR #4151, opened 2020-07-30
Automatically load all entries listed in a central JSON

malte0811:
Demo pack adding an entry in a new category: [pack.zip](https://github.com/BluSunrize/ImmersiveEngineering/files/5002172/pack.zip) (I hope you'll excuse me not uploading that to pastebin)

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/client/IEManual.java
+++ b/src/main/java/blusunrize/immersiveengineering/client/IEManual.java
@@ -99,22 +99,14 @@ public class IEManual
 		IEManualInstance ieMan = ManualHelper.getManual();
 		InnerNode<ResourceLocation, ManualEntry> generalCat = ieMan.getRoot().getOrCreateSubnode(new ResourceLocation(MODID,
 				ManualHelper.CAT_GENERAL), 0);
-		InnerNode<ResourceLocation, ManualEntry> constructionCat = ieMan.getRoot().getOrCreateSubnode(new ResourceLocation(MODID,
-				ManualHelper.CAT_CONSTRUCTION), 10);
+		ieMan.getRoot().getOrCreateSubnode(new ResourceLocation(MODID, ManualHelper.CAT_CONSTRUCTION), 10);
 		InnerNode<ResourceLocation, ManualEntry> energyCat = ieMan.getRoot().getOrCreateSubnode(new ResourceLocation(MODID,
 				ManualHelper.CAT_ENERGY), 20);
-		InnerNode<ResourceLocation, ManualEntry> toolsCat = ieMan.getRoot().getOrCreateSubnode(new ResourceLocation(MODID,
-				ManualHelper.CAT_TOOLS), 30);
-		InnerNode<ResourceLocation, ManualEntry> machinesCat = ieMan.getRoot().getOrCreateSubnode(new ResourceLocation(MODID,
-				ManualHelper.CAT_MACHINES), 40);
+		ieMan.getRoot().getOrCreateSubnode(new ResourceLocation(MODID, ManualHelper.CAT_TOOLS), 30);
+		ieMan.getRoot().getOrCreateSubnode(new ResourceLocation(MODID, ManualHelper.CAT_MACHINES), 40);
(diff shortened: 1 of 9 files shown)


src/main/java/blusunrize/immersiveengineering/client/IEManual.java
```diff
@@ -124,14 +113,8 @@ public static void addIEManualEntries()
 					)
 			);
 			builder.readFromFile(new ResourceLocation(MODID, "thermoelectric"));
-			ieMan.addEntry(energyCat, builder.create());
+			ieMan.addEntry(energyCat, builder.create(), ieMan.atOffsetFrom(energyCat, "redstone_wire", 0.5));
```
BluSunrize, line 116: I appreciate the idea of using an "offset from" stategy, but what if someone completely replaces the manual's contents? Won't this throw an exception?

src/main/java/blusunrize/lib/manual/Tree.java
```diff
@@ -143,11 +161,32 @@ public NT getNodeData()
...
 
 		public InnerNode<NT, LT> getOrCreateSubnode(NT data, int weight)
+		{
+			return getOrCreateSubnode(data, () -> weight);
+		}
+
+		public InnerNode<NT, LT> getOrCreateSubnode(NT data, DoubleSupplier weight)
+		{
+			return getSubnode(data)
+					.orElseGet(() -> addNewSubnode(data, weight));
```
BluSunrize, line 171: This line seems short and simple enough to not require a break. Any reason here, something I should improve about my style as well?

src/main/resources/assets/immersiveengineering/manual/autoload.json
```diff
@@ -0,0 +1,99 @@
...
+      "barrel",
+      "metalconstruction",
+      "metal_barrel",
+      "concrete",
+      "lighting",
+      "silo",
+      "tank"
+    ]
+  },
+  "energy": {
```
BluSunrize, line 38: Why does this category not have a weight? Same for heavy machines?

BluSunrize (changes requested):

src/main/java/blusunrize/immersiveengineering/client/IEManual.java
malte0811: > what if someone completely replaces the manual's contents?

How would they? I'm loading *all* files at that location in active resource packs, so the only reason why that entry would be missing would be if the IE resource pack is disabled. And I suspect that a lot of other things would break in that situation.

That has reminded me of an issue I need to look into though, there's no guarantee for the order of the `autoload` files from different packs as far as I know. So autoloaded entries from IE might end up below modded entries on one launch, but above on the next launch. The best approach would probably be to have a `autoloadPriority` or similar key at the top level of the JSON and load them in increasing order of that property.

src/main/java/blusunrize/lib/manual/Tree.java
malte0811: I think this may have been more complex before, and I forgot to remove the line break after simplifying it. So no real reason.

src/main/resources/assets/immersiveengineering/manual/autoload.json
malte0811: Because those categories are created in code (and long before the JSON is loaded), so the weight in the JSON would just be ignored. I didn't want to put a value in here to prevent confusion along the lines of "I changed this value, but the order in the manual didn't change!".

src/main/java/blusunrize/immersiveengineering/client/IEManual.java
BluSunrize: Ah, I missed that part then. I assumed a resource pack was overwriting the file, not appending it.

src/main/resources/assets/immersiveengineering/manual/autoload.json
BluSunrize: I see, fair enough. I suppose it would be nice if all categories loaded from this file for the sake of readability, but I recognize why order of operations does not allow for this.

BluSunrize (approved):
LGTM

BluSunrize:
minus the merge conflicts, of course...

PR #4182, opened 2020-08-08
Add waterlogging of blocks

BluSunrize:
Seems to be working well so far.
Did I miss any relevant ones?

- Scaffolding
- Wallmounts
- Posts
- Structural Arms
- Lanterns
- Balloons
- Shader Banners
- Connectors & Relays
- Transformers
- Breakers
- Razor Wire
- Feedthrough Connectors
- Conveyors
- Pipes

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
@@ -307,18 +307,18 @@ public class IEBaseBlock extends Block implements IIEBlock, IWaterLoggable
 	{
 		ItemStack activeStack = player.getHeldItem(hand);
 		if(activeStack.getToolTypes().contains(HammerItem.HAMMER_TOOL))
-			return hammerUseSide(hit.getFace(), player, world, pos, hit);
+			return hammerUseSide(hit.getFace(), player, hand, world, pos, hit);
 		if(activeStack.getToolTypes().contains(ScrewdriverItem.SCREWDRIVER_TOOL))
-			return screwdriverUseSide(hit.getFace(), player, world, pos, hit);
+			return screwdriverUseSide(hit.getFace(), player, hand, world, pos, hit);
 		return super.onBlockActivated(state, world, pos, player, hand, hit);
 	}
 
-	public boolean hammerUseSide(Direction side, PlayerEntity player, World w, BlockPos pos, BlockRayTraceResult hit)
+	public boolean hammerUseSide(Direction side, PlayerEntity player, Hand hand, World w, BlockPos pos, BlockRayTraceResult hit)
 	{
 		return false;
 	}
 
-	public boolean screwdriverUseSide(Direction side, PlayerEntity player, World w, BlockPos pos, BlockRayTraceResult hit)
+	public boolean screwdriverUseSide(Direction side, PlayerEntity player, Hand hand, World w, BlockPos pos, BlockRayTraceResult hit)
 	{
 		return false;
 	}
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
@@ -13,39 +13,44 @@ import blusunrize.immersiveengineering.common.IEContent;
 import blusunrize.immersiveengineering.common.items.HammerItem;
 import blusunrize.immersiveengineering.common.items.ScrewdriverItem;
 import blusunrize.immersiveengineering.common.items.WirecutterItem;
-import blusunrize.immersiveengineering.common.util.IELogger;
 import com.google.common.base.Preconditions;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
+import net.minecraft.block.IWaterLoggable;
 import net.minecraft.block.material.PushReaction;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.LivingEntity;
 import net.minecraft.entity.player.PlayerEntity;
-import net.minecraft.item.*;
-import net.minecraft.item.Item.Properties;
+import net.minecraft.fluid.Fluid;
+import net.minecraft.fluid.Fluids;
+import net.minecraft.fluid.IFluidState;
+import net.minecraft.item.BlockItemUseContext;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemGroup;
+import net.minecraft.item.ItemStack;
 import net.minecraft.state.IProperty;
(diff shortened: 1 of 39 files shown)


BluSunrize:
@malte0811 This is ready for review, if you have some time.

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
```diff
@@ -218,7 +223,10 @@ protected void fillStateContainer(Builder<Block, BlockState> builder)
 
 	protected BlockState getInitDefaultState()
 	{
-		return this.stateContainer.getBaseState();
+		BlockState state = this.stateContainer.getBaseState();
+		if(state.has(BlockStateProperties.WATERLOGGED))
+			state = state.with(BlockStateProperties.WATERLOGGED, Boolean.FALSE);
```
malte0811, line 228: Why does this need to be set to false explicitly? Vanilla doesn't seem to do that either.

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
```diff
@@ -299,22 +307,75 @@ public boolean onBlockActivated(BlockState state, World world, BlockPos pos, Pla
...
+
+	@Override
+	public BlockState getStateForPlacement(BlockItemUseContext context)
+	{
+		BlockState state = this.getDefaultState();
+		if(state.has(BlockStateProperties.WATERLOGGED)
+				&&context.getWorld().getFluidState(context.getPos()).getFluid()==Fluids.WATER)
+		{
+			state = state.with(BlockStateProperties.WATERLOGGED, true);
+		}
```
malte0811: Formatting (extra `{}`)

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
```diff
@@ -299,22 +307,75 @@ public boolean onBlockActivated(BlockState state, World world, BlockPos pos, Pla
...
+		return state;
+	}
+
+	@Override
+	public BlockState updatePostPlacement(BlockState stateIn, Direction facing, BlockState facingState, IWorld worldIn, BlockPos currentPos, BlockPos facingPos)
+	{
+		if (stateIn.has(BlockStateProperties.WATERLOGGED) && stateIn.get(BlockStateProperties.WATERLOGGED)) {
+			worldIn.getPendingFluidTicks().scheduleTick(currentPos, Fluids.WATER, Fluids.WATER.getTickRate(worldIn));
+		}
```
malte0811: Same, this time even with wrong placement. I assume this call is based on vanilla code?

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
```diff
@@ -13,39 +13,44 @@
...
-import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.function.BiFunction;
-import java.util.function.Function;
 
-public class IEBaseBlock extends Block implements IIEBlock
+public class IEBaseBlock extends Block implements IIEBlock, IWaterLoggable
```
malte0811, line 53: Does making all IE blocks extend `IWaterLoggable` cause issues anywhere? Or does everything actually check that `canContainFluid` is true?

src/main/java/blusunrize/immersiveengineering/common/blocks/generic/PostBlock.java
```diff
@@ -73,26 +76,29 @@ public PostBlock(String name, Properties blockProps)
...
+					BlockState armState = world.getBlockState(armPos);
+					if(armState.getBlock()==this&&armState.get(HORIZONTAL_OFFSET).getOffset().equals(d.getDirectionVec()))
+						world.setBlockState(armPos, Blocks.AIR.getDefaultState());
+				}
+				for(int i = 0; i <= highestBlock; ++i)
+					world.setBlockState(pos.up(i), Blocks.AIR.getDefaultState());
 			}
-			for(int i = 0; i <= highestBlock; ++i)
-				world.setBlockState(pos.up(i), Blocks.AIR.getDefaultState());
+			super.onReplaced(state, world, pos, newState, moving);
```
malte0811: Should this be inside the `if`?

src/main/java/blusunrize/immersiveengineering/common/items/HammerItem.java
```diff
@@ -171,7 +171,7 @@ public ActionResultType onItemUseFirst(ItemStack stack, ItemUseContext context)
 				return ActionResultType.SUCCESS;
 			else if(!rotate&&tile instanceof IHammerInteraction)
 			{
-				if(((IHammerInteraction)tile).hammerUseSide(side, player, context.getHitVec()))
+				if(((IHammerInteraction)tile).hammerUseSide(side, player, context.getHand(), context.getHitVec()))
```
malte0811, line 174: Any particular reason for this change? I don't think I saw the new parameter used anywhere.

malte0811 (changes requested):

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
BluSunrize: Vanilla does this on the method that fills the state. We set any property to their first possible value (which is why we have a BooleanInverted property). The first value on Mojang's boolean properties is true (disgustingly enough. 0 should be false), so I had to explicitely set it to false here.

src/main/java/blusunrize/immersiveengineering/common/blocks/IEBaseBlock.java
BluSunrize: It does from as far as I could see. I checked uses of IWaterLoggable and found nothing that should present an issue.

src/main/java/blusunrize/immersiveengineering/common/blocks/generic/PostBlock.java
BluSunrize: Doesn't *really* matter, because the method it overrides has another block != block check in it. It comes out to the same result

src/main/java/blusunrize/immersiveengineering/common/items/HammerItem.java
BluSunrize: Posts. They needed to create an ItemUseContext to make an isReplacable check. That context needs a hand, which wasn't being passed down.
Also, this may come in useful in the future.

PR #4199, opened 2020-08-11
Adjust version numbering to be semantic

BluSunrize:
And do so especially for Jenkins builds.


BluSunrize:
Merging this immediately, since we already talked about it in Discord.

Issue #4117, opened 2020-07-21
[1.14+] Use Tags for Fluids allowed in the Diesel Generator (and maybe Chemthrower)


BluSunrize:
This was pointed out by KnightMiner, who educated me on Fluids using namespaced registries now, and allowing for Tags.
That in mind, our DieselGen (and maybe the Chemthrower) should identify its usable fluids based on tags, not IDs.

That also entails actually *applying* those tags to our fluids.

This is a ToDo for coming releases.

PR #4225, opened 2020-08-17
Refactor: Fluid Tags as inputs

BluSunrize:
It started with Diesel Gen and Chemthrower (see #4117), but then pupnewfster pointed out that they do all fluid inputs in Mekanism based on tags.
The reason for that is that all fluids are namespaced now, so `immersiveengineering:ethanol` is different than `<other mod>:ethanol`

Now we have a class that stores a Fluid tag, an amount, and optional some NBT data (necessary for potion mixing & bottling) and we use those for anything that uses Fluid *inputs* (Mixer, Bottling, Refinery). Outputs have remained as FluidStacks, because implementing our preference system here would be too much of a mess.


BluSunrize:
Malte, with the changes made to the API here, I believe that puts us at release version 4.0.0, correct?

src/main/java/blusunrize/immersiveengineering/api/energy/DieselHandler.java
```diff
@@ -24,66 +24,52 @@
  */
 public class DieselHandler
 {
-	static final HashMap<ResourceLocation, Integer> dieselGenBurnTime = new HashMap<>();
-	static final Set<Fluid> drillFuel = new HashSet<>();
+	static final HashMap<Tag<Fluid>, Integer> dieselGenBurnTime = new HashMap<>();
```
malte0811: Can we change the type of the field to `Map<…>` instead of `HashMap<…>`? I'm not sure whether we actually still want a `Map` here in any case, since we only ever iterate over it. Maybe `List<Pair<…>>` would be more appropriate?

src/main/java/blusunrize/immersiveengineering/api/tool/ChemthrowerHandler.java
```diff
@@ -33,100 +37,48 @@
...
 import java.util.HashSet;
+import java.util.Map;
 
 public class ChemthrowerHandler
 {
-	public static HashMap<ResourceLocation, ChemthrowerEffect> effectMap = new HashMap<>();
-	public static HashSet<ResourceLocation> flammableList = new HashSet<>();
-	public static HashSet<ResourceLocation> gasList = new HashSet<>();
+	public static HashMap<Tag<Fluid>, ChemthrowerEffect> effectMap = new HashMap<>();
+	public static HashSet<Tag<Fluid>> flammableList = new HashSet<>();
```
malte0811, line 46: Same here

src/main/java/blusunrize/immersiveengineering/api/crafting/FluidTagInput.java
```diff
@@ -0,0 +1,145 @@
...
+		ResourceLocation resourceLocation = new ResourceLocation(JSONUtils.getString(jsonObject, "tag"));
+		if(!JSONUtils.hasField(jsonObject, "nbt"))
+			return new FluidTagInput(resourceLocation, JSONUtils.getInt(jsonObject, "amount"));
+		try
+		{
+			CompoundNBT nbt = JsonToNBT.getTagFromJson(JSONUtils.getString(jsonObject, "nbt"));
+			return new FluidTagInput(resourceLocation, JSONUtils.getInt(jsonObject, "amount"), nbt);
+		} catch(CommandSyntaxException e)
+		{
+			throw new RuntimeException(e);
```
SkySomDynata, line 70: RuntimeExceptions are bad in deserialize use IllegalArgument or JsonParse Exceptions

src/main/java/blusunrize/immersiveengineering/api/crafting/FluidTagInput.java
```diff
@@ -0,0 +1,145 @@
...
+	public static FluidTagInput deserialize(JsonElement input)
+	{
+		Preconditions.checkArgument(input instanceof JsonObject, "FluidTagWithSize can only be deserialized from a JsonObject");
+		JsonObject jsonObject = input.getAsJsonObject();
+		ResourceLocation resourceLocation = new ResourceLocation(JSONUtils.getString(jsonObject, "tag"));
+		if(!JSONUtils.hasField(jsonObject, "nbt"))
+			return new FluidTagInput(resourceLocation, JSONUtils.getInt(jsonObject, "amount"));
+		try
+		{
+			CompoundNBT nbt = JsonToNBT.getTagFromJson(JSONUtils.getString(jsonObject, "nbt"));
```
SkySomDynata, line 66: Somethinglike this (Pulled from CraftingHelper) would allow for users doing packs to not have to write an escape NBT string (Yes I know for Datagens it's not awful)
```                
if(element.isJsonObject())
  nbt = JsonToNBT.getTagFromJson(GSON.toJson(element));
else
  nbt = JsonToNBT.getTagFromJson(JSONUtils.getString(element, "nbt"));
```

malte0811 (changes requested):

src/main/java/blusunrize/immersiveengineering/api/crafting/FluidTagInput.java
SkySom: Use IllegalArgumentException or JsonParseException as those are what RecipeManager catches.

src/main/java/blusunrize/immersiveengineering/api/crafting/FluidTagInput.java
SkySom: Somethinglike this (Pulled from CraftingHelper) would allow for users doing packs to not have to write an escape NBT string (Yes I know for Datagens it's not awful)
```                
if(element.isJsonObject())
  nbt = JsonToNBT.getTagFromJson(GSON.toJson(element));
else
  nbt = JsonToNBT.getTagFromJson(JSONUtils.getString(element, "nbt"));
```

SkySom (changes requested):

src/main/java/blusunrize/immersiveengineering/api/energy/DieselHandler.java
BluSunrize: Yeah we are using it like a list, you make a good point.

src/main/java/blusunrize/immersiveengineering/api/crafting/FluidTagInput.java
BluSunrize: Fair point. I was matching vanilla handling here.

BluSunrize:
Ready for another review

SkySom (approved):

malte0811 (approved):

PR #4258, opened 2020-08-28
Split multiblock models

malte0811:
This has two advantages over the old approach of a single model rendered by one block:
1. Chunk culling will no longer make IE machines vanish under certain angles. Wires are still affected by this, but that would require a totally different fix.
2. Lighting should be a lot more consistent, at least when the "Forge experimental lighting pipeline" is enabled (most lighting code is not intended to deal with models larger than a single block).

The PR looks a lot bigger than it actually is due to changes to generated data.


src/main/java/blusunrize/immersiveengineering/client/models/CompositeBakedModel.java
```diff
@@ -0,0 +1,98 @@
+package blusunrize.immersiveengineering.client.models;
```
BluSunrize, line 10: Missing license comment?

src/main/java/blusunrize/immersiveengineering/api/client/ICacheKeyProvider.java
```diff
@@ -0,0 +1,25 @@
+package blusunrize.immersiveengineering.api.client;
```
BluSunrize, line 10: Missing license comment?

src/main/java/blusunrize/immersiveengineering/api/client/IModelOffsetProvider.java
```diff
@@ -0,0 +1,13 @@
+package blusunrize.immersiveengineering.api.client;
```
BluSunrize, line 10: Missing license comment?

src/main/java/blusunrize/immersiveengineering/client/models/obj/IESmartObjModel.java
```diff
@@ -321,29 +322,52 @@ public IModelData getModelData(@Nonnull ILightReader world, @Nonnull BlockPos po
...
+			tex = extraData.getData(Model.TEXTURE_REMAP);
+		return getKey(state, extraData, true, objState, tex);
+	}
+
+	public RenderCacheKey getKey(
+			@Nullable BlockState blockState,
+			@Nonnull IModelData modelData,
+			boolean addAnimationAndTex,
+			IEObjState visibility,
+			Map<String, String> tex
```
BluSunrize: This feels like pretty aggressive linebreaking. Especially when the method above doesn't do that. My cutoff point is at about 120 characters.

src/main/java/blusunrize/immersiveengineering/client/models/split/BakedBasicSplitModel.java
```diff
@@ -0,0 +1,133 @@
...
+			return base.getQuads(state, side, rand, extraData);
+	}
+
+	@Nonnull
+	@Override
+	public IModelData getModelData(
+			@Nonnull ILightReader world,
+			@Nonnull BlockPos pos,
+			@Nonnull BlockState state,
+			@Nonnull IModelData tileData
```
BluSunrize: See previous comment on linebreaking.

src/main/java/blusunrize/immersiveengineering/client/models/split/PolygonUtils.java
```diff
@@ -0,0 +1,130 @@
+package blusunrize.immersiveengineering.client.models.split;
```
BluSunrize, line 10: Missing license comment?

src/main/java/blusunrize/immersiveengineering/client/models/split/BakedDynamicSplitModel.java
```diff
@@ -0,0 +1,101 @@
+package blusunrize.immersiveengineering.client.models.split;
```
BluSunrize, line 10: Missing license comment?

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/ConveyorBeltTileEntity.java
```diff
@@ -237,6 +240,12 @@ public boolean isRSPowered()
 		return super.isRSPowered();
 	}
 
+	@Override
+	public BlockPos getModelOffset(BlockState state)
+	{
+		throw new UnsupportedOperationException();
+	}
+
```
BluSunrize: Why are conveyors implementing this at all?

BluSunrize (changes requested):
Looks pretty good to me, bar one question and some missing comments.

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/ConveyorBeltTileEntity.java
malte0811: I originally added `getModelOffset` to `IGeneralMultiblock`, which conveyors seem to implement. And when I moved it to a different interface I accidentally added it here instead of removing the implementation.

src/main/java/blusunrize/immersiveengineering/client/models/obj/IESmartObjModel.java
malte0811: It's a habit that came over from work, there our rule is to either put everything in the same line or one parameter per line. I'll fix it later today.

src/main/java/blusunrize/immersiveengineering/api/client/ICacheKeyProvider.java
malte0811: Hm, not sure what's causing that. At one point I intentionally removed a comment (since the file was part of BMS), but that doesn't explain why it's missing everywhere else.

BluSunrize (approved):

SkySom (approved):

PR #4261, opened 2020-08-29
Add option to not handle Last Active on Manual Instance

SkySom:
This should allow opening other versions of the Manual without overwriting.

also includes getter for IMultiblock

### DIFF
--- a/src/main/java/blusunrize/lib/manual/gui/ManualScreen.java
+++ b/src/main/java/blusunrize/lib/manual/gui/ManualScreen.java
@@ -55,14 +55,14 @@ public class ManualScreen extends Screen
 	private ClickableList entryList;
 	private ClickableList suggestionList;
 
-	private final boolean doLastActive;
+	private final boolean setLastActive;
 
 	public ManualScreen(ManualInstance manual, String texture)
 	{
 		this(manual, texture, true);
 	}
 
-	public ManualScreen(ManualInstance manual, String texture, boolean doLastActive)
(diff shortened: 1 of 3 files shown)


src/main/java/blusunrize/lib/manual/gui/ManualScreen.java
```diff
@@ -154,7 +163,11 @@ else if(currentNode.getSuperNode()!=null)
 		}
 		else if(searchField!=null)
 			searchField = null;
-		lastActiveManual = this;
+
+		if (doLastActive)
+		{
+			lastActiveManual = this;
+		}
```
malte0811: Formatting (no baces on single-line `if`s)

src/main/java/blusunrize/lib/manual/gui/ManualScreen.java
```diff
@@ -55,7 +55,14 @@
 	private ClickableList entryList;
 	private ClickableList suggestionList;
 
+	private final boolean doLastActive;
```
malte0811: I would have called it something different here (`setAsLastActive` or similar), but this works as well.

malte0811 (changes requested):

src/main/java/blusunrize/lib/manual/gui/ManualScreen.java
BluSunrize: I agree, naming here is a bit awkward.

BluSunrize (approved):
LGTM

malte0811 (approved):

PR #4264, opened 2020-08-30
Add overridable methods for IEBaseBlock

p3lim:
Adding these methods gives addons for IE the ability to re-use classes that extends IEBaseBlock (and IEBaseBlock itself) without having to resort to hacks to get the block and blockitem out of IE. This was already half-way there with the `createRegistryName()` method.

I decided to separate the block and blockitem registration boolean methods just in case, but they could just as well be merged into one method, as if you're registering one with IE you'll probably want to register both anyways.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/BlockItemIE.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/BlockItemIE.java
@@ -42,7 +42,7 @@ public class BlockItemIE extends BlockItem
 
 	public BlockItemIE(Block b)
 	{
-		this(b, new Item.Properties().group(ImmersiveEngineering.itemGroup));
+		this(b, new Item.Properties().group(b.getItemGroup()));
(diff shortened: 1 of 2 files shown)


malte0811:
1. This PR does not seem to be compileable (unless the github action has a bug).
2. Using `IEBaseBlock` in addons is not recommended. I did it when I started Industrial Wires, but (IIRC) moved away from it at some point. The class is not part of the API, so breaking changes can and will happen at any time. What parts do you actually need? I would recommend writing your own base block class that fits your needs and that you can modify as necessary.

p3lim:
I'm implementing a new connector, which has a lot of logic to it that I'd prefer not to reimplement since it's identical.
Both the wires and tile entities have APIs, only the block remains.
Being new to modding this seemed like a good option, but I might have been wrong.

BluSunrize:
You shouldn't need this block either. Connectors depend on TileEntities, the logic for which is already in the API. You should be able to use your own blocks to house these TEs.

PR #4271, opened 2020-09-03
Metal press rendersync

MalkContent:
gets rid of a lot of magic numbers in MetalPressRenderer and MetalPressTileEntity and centralizes them in MetalPressRenderer to allow for easier manipulation of the machines operating speed/behavior

added code to automatically speed up the transport and press speeds for processes with maxTicks smaller than the standard operating speed allow by default

removed usage of partialTicks because of at seemingly randomly timed jaggy animation of the items on the conveyor belt, more noticably at faster processing speeds
sadly could not find a solution to make it use partialTicks properly :/


src/main/java/blusunrize/immersiveengineering/client/render/tile/MetalPressRenderer.java
```diff
@@ -59,30 +63,33 @@ public void render(MetalPressTileEntity te, double x, double y, double z, float
 		GlStateManager.translated(x+.5, y+.5, z+.5);
 		float piston = 0;
 		float[] shift = new float[te.processQueue.size()];
+		System.out.println(partialTicks);
```
SkySom: println needs to be removed.

SkySom (changes requested):

src/main/java/blusunrize/immersiveengineering/client/render/tile/MetalPressRenderer.java
MalkContent: ty, good catch

SkySom (approved):

BluSunrize:
This should be obsolete due to #4272, correct?

MalkContent:
correct, it's included in #4272

PR #4272, opened 2020-09-03
Metal press conveyor sync

MalkContent:
made transport time dependent on conveyor animation - to visibly transport items with the conveyor belt instead of them sliding along at different speeds

halved processTime to better fit the animation, indirectly buffing the MetalPress. I don't insist on this, because I generally don't want to just throw in stuff that could affect balance. I'd say just look at how it runs for yourself


BluSunrize:
So what about #4271 ?

This seems to include it. Can you explain to me which PR I should be looking at?

BluSunrize:
Also, why is this posed against the 1.14 branch when I changed our default branch to 1.15 last week? :^)

MalkContent:
Made it separate because the other one is morejust a little bit of code cleanup and this one has stuff in it that can affect balance, but it builds on the other one. So in case you don't like this one, you could still take the other :)

MalkContent:
1.14 because I had this in the work for some time, an just slow. And rushed it out when you made the announcement ^^;
Should I port this to 1.15?

BluSunrize:
That would be super great!
I like this, because increasing the speed of the press actually feels good. I like the balance change, and it coincides with visual improvements, so win/win

If you throw this against 1.15, that would be much appreciated ^_^
Just make sure to make your constants in the renderer use capital letters (that would have been part of my review)

MalkContent:
alright. will do :)

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/MetalPressTileEntity.java
```diff
@@ -164,7 +166,9 @@ public void onEntityCollision(World world, Entity entity)
 			if(recipe==null)
 				return;
 			ItemStack displayStack = recipe.getDisplayStack(stack);
-			float transformationPoint = 56.25f/120f;
+//			float processMaxTicks = recipe.getTotalProcessTime();
+//			float transformationPoint = (MetalPressRenderer.getPressTime(processMaxTicks)+MetalPressRenderer.getPressTime(processMaxTicks))/processMaxTicks;
+			float transformationPoint = 0.5f;
```
BluSunrize, line 171: Why's there commented out code here?

src/main/java/blusunrize/immersiveengineering/client/render/tile/MetalPressRenderer.java
```diff
@@ -34,6 +34,11 @@
 {
 	public static DynamicModel<Void> PISTON;
 
+	private static final float translationDistance = 2.5f;
+	private static final float standardTransportTime = 16f*(translationDistance/2); //16 frames in conveyor animation, 1 frame/tick, 2.5 blocks of total translation distance, halved because transport time just affects half the distance
+	private static final float standardPressTime = 3.75f;
+	private static final float minCycleTime = 60f; //set >= 2*(standardPressTime+standardTransportTime)
+
```
BluSunrize, line 41: Constants in capital letters please.

BluSunrize (changes requested):
LGTM, just some minor changes please.

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/MetalPressTileEntity.java
MalkContent: I left it there in case there's a reason to put the transformationPoint to the moment where the press piston hits the conveyor
I put in 0.5f because it required no calculation and calls and still works for all recipes

I can remove/change it if you want

src/main/java/blusunrize/immersiveengineering/client/render/tile/MetalPressRenderer.java
MalkContent: good point, done

BluSunrize (approved):
LGTM

PR #4289, opened 2020-09-10
Generated recipe lists

malte0811:
Sync generated recipes instead of generating them on both server and client. This PR also moves mixer potion recipes to use the standard class, to simplify code and fix JEI display of those recipes.


src/main/java/blusunrize/immersiveengineering/common/crafting/ArcRecyclingThreadHandler.java
```diff
@@ -41,6 +44,7 @@ public ArcRecyclingThreadHandler(Collection<IRecipe<?>> allRecipes)
 		this.recipeList = allRecipes.stream()
 				.filter(ArcFurnaceRecipe.assembleRecyclingFilter())
 				.collect(Collectors.toList());
+		runningHandler = this;
```
BluSunrize, line 47: Could have kept style here and put a `this.` before it

src/main/java/blusunrize/immersiveengineering/common/crafting/serializers/GeneratedListSerializer.java
```diff
@@ -0,0 +1,79 @@
...
+	public void write(@Nonnull PacketBuffer buffer, @Nonnull GeneratedListRecipe recipe)
+	{
+		List<? extends IESerializableRecipe> recipes = recipe.getSubRecipes();
+		buffer.writeVarInt(recipes.size());
+		buffer.writeResourceLocation(recipe.getSubSerializer());
+		for(IESerializableRecipe r : recipes)
+		{
+			buffer.writeResourceLocation(r.getId());
+			((IERecipeSerializer)r.getSerializer()).write(buffer, r);
+		}
```
BluSunrize, line 77: Should we be concerned about exceeding packet limits with this solution?

BluSunrize (commented):
Minor change request, big questionmark on the entire idea.

src/main/java/blusunrize/immersiveengineering/common/crafting/ArcRecyclingThreadHandler.java
malte0811: No, because `runningHandler` is static. I can put `ArcRecyclingThreadHandler.` before it if you want.

src/main/java/blusunrize/immersiveengineering/common/crafting/serializers/GeneratedListSerializer.java
malte0811: As far as I can tell all recipes are batched into a single packet anyway, so it should be fine/should not make things much worse than they already are in modpacks.

src/main/java/blusunrize/immersiveengineering/common/crafting/ArcRecyclingThreadHandler.java
BluSunrize: That checks out, I'm a dumb dumb!

src/main/java/blusunrize/immersiveengineering/common/crafting/serializers/GeneratedListSerializer.java
BluSunrize: Fair!

BluSunrize (approved):
LGTM

SkySom (approved):

PR #4310, opened 2020-09-20
Use VBOs for some tile renders

malte0811:
Also changes some other TERs to use `ClientUtils#renderModelTESRFast` as it's faster than the block model render and looks very similar in those cases. The crusher actually looks better now IMO: Before one side of the barrels was usually darker than the other, but as the rotation wasn't included in the lighting the darker side rotated with the barrels. Now all sides have the same, roughly correct brightness.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/api/client/IVertexBufferHolder.java
+++ b/src/main/java/blusunrize/immersiveengineering/api/client/IVertexBufferHolder.java
@@ -1,0 +1,40 @@
+/*
+ * BluSunrize
+ * Copyright (c) 2020
+ *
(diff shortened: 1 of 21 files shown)


BluSunrize (approved):
It looks fine to me, as far as I can tell.

I remember you mentioning that the coremod is a workaround? Are we likely to see a hook in forge in the future?

malte0811:
There's a Forge PR for this as well as some other, related events: https://github.com/MinecraftForge/MinecraftForge/pull/7225. You'll need to decide for yourself whether that's "likely" to be merged in the future.

PR #4320, opened 2020-09-22
Add the Sawmill

BluSunrize:
It stripes logs, then cuts them into planks.
I think I covered everything in there, changelog comes after merge.

Still missing a subtitle for the manual, but that can come later. Also hopefully getting a better sawdust texture from Hazard.

I did find a bug with IngredientFluidStack and shapeless recipes, which is fixed in:
https://github.com/BluSunrize/ImmersiveEngineering/commit/22731efa28d131df7765e70bdab32dd35cd199da
We need to monitor if that fix breaks on 1.16+ due to the timing at which tags load.


malte0811:
> I did find a bug with IngredientFluidStack and shapeless recipes, which is fixed in: 22731ef

Can you elaborate on that? I'm pretty sure we need to override `isSimple` as "simple" means "only cares about the item ID". The `hasNoMatchingItems` looks reasonable, though I'd expect the default implementation to be something like `getMatchingStacks().size() == 0`, which should work here.

src/main/java/blusunrize/immersiveengineering/client/render/tile/SawmillRenderer.java
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	@Override
+	public void render(SawmillTileEntity te, float partialTicks, MatrixStack matrixStack, IRenderTypeBuffer bufferIn, int combinedLightIn, int combinedOverlayIn)
+	{
+		if(!te.formed||te.isDummy()||!te.getWorldNonnull().isBlockLoaded(te.getPos()))
+			return;
+
+		//Grab model
+		final BlockRendererDispatcher blockRenderer = Minecraft.getInstance().getBlockRendererDispatcher();
```
malte0811: Unused variable?

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/SawmillTileEntity.java
```diff
@@ -0,0 +1,596 @@
...
+		super.tick();
+		if(isDummy()||isRSDisabled())
+			return;
+		if(world.isRemote&&shouldRenderAsActive())
+		{
+			animation_bladeRotation += 36f;
+			animation_bladeRotation %= 360f;
+		}
+		if(world.isRemote)
+			return;
```
malte0811, line 151: I would prefer something like
```java
if (world.isRemove)
{
   if (shouldRenderAsActive())
   {…}
   return;
}
```
but this is fine too.

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/SawmillTileEntity.java
```diff
@@ -0,0 +1,596 @@
...
+		for(ItemStack output : secondaries)
+			doSecondaryOutput(output.copy());
+	}
+
+	@Override
+	public boolean interact(Direction side, PlayerEntity player, Hand hand, ItemStack heldItem, float hitX, float hitY, float hitZ)
+	{
+		SawmillTileEntity master = master();
+		if(master!=null)
+			if(player.isSneaking()&&!master.sawblade.isEmpty())
```
malte0811, line 187: I've said this before, I don't really like `if(a) if(b) … else` without braces on the outer if. Could you add those?

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/SawmillTileEntity.java
```diff
@@ -0,0 +1,596 @@
...
+					player.setHeldItem(hand, master.sawblade.copy());
+				else if(!world.isRemote)
+					player.entityDropItem(master.sawblade.copy(), 0);
+				master.sawblade = ItemStack.EMPTY;
+				this.updateMasterBlock(null, true);
+				return true;
+			}
+			else if(IETags.sawblades.contains(heldItem.getItem()))
+			{
+				ItemStack tempMold = !master.sawblade.isEmpty()?master.sawblade.copy(): ItemStack.EMPTY;
```
malte0811: I would guess you copied this from the metal press and forgot to rename it?

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/SawmillTileEntity.java
```diff
@@ -0,0 +1,596 @@
...
+				master.sawblade = Utils.copyStackWithAmount(heldItem, 1);
+				heldItem.shrink(1);
+				if(heldItem.getCount() <= 0)
+					heldItem = ItemStack.EMPTY;
+				else
+					player.setHeldItem(hand, heldItem);
+				if(!tempMold.isEmpty())
+					if(heldItem.isEmpty())
```
malte0811, line 208: Same comment on `if(a)if(b) else` as before.

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/SawmillTileEntity.java
```diff
@@ -0,0 +1,596 @@
...
+			SawmillTileEntity master = master();
+			if(master==null)
+				return;
+			ItemStack stack = ((ItemEntity)entity).getItem();
+			if(stack.isEmpty())
+				return;
+			master.insertItemToProcess(stack, false);
+			if(stack.getCount() <= 0)
+				entity.remove();
+		}
```
malte0811, line 384: I'm disappointed, stepping on the giant exposed spinning sawblade does not hurt you?

malte0811 (changes requested):

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/SawmillTileEntity.java
BluSunrize: ...you make a good point.

BluSunrize:
Ready for another review

malte0811 (approved):
Looks good now.

PR #4322, opened 2020-09-23
Bottling machine lift fix

### LINKED ISSUE #4313: [MC1.15 | IE119][minor] Bottling Machine render inconsistency
The bottling machine's lifting fork is set below the conveyor level when not in use (see picture).
However, at the start and finish of the lifting animation, the fork jumps from being set below to conveyorlevel and vice versa, instead of lifting from where it is resting. Don't know if the resting position is  wrong or the animation's border-position, but there you have it.
grafik

MalkContent:
fixes the animation jump
also changed the animation a little to make the items on the fork less hover-y, but if you don't like that, just ignore the second commit

fixes: https://github.com/BluSunrize/ImmersiveEngineering/issues/4313

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/client/render/tile/BottlingMachineRenderer.java
+++ b/src/main/java/blusunrize/immersiveengineering/client/render/tile/BottlingMachineRenderer.java
@@ -103,8 +103,8 @@ public class BottlingMachineRenderer extends TileEntityRenderer<BottlingMachineT
 					lift = (85-processTimer)/20f*.125f;
 					itemFill = 1;
 				}
-				if (lift > pixelHeight)
-				itemY += lift-pixelHeight;
+				if(lift > pixelHeight)
+					itemY += lift-pixelHeight;
 			}
 			else
 			{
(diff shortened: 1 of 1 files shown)


src/main/java/blusunrize/immersiveengineering/client/render/tile/BottlingMachineRenderer.java
```diff
@@ -102,8 +103,8 @@ else if(processTimer <= 65)
 					lift = (85-processTimer)/20f*.125f;
 					itemFill = 1;
 				}
-				itemY += lift;
-				lift += .0625;
+				if (lift > pixelHeight)
+				itemY += lift-pixelHeight;
```
SkySom: Is this supposed to be part of the if? should be indented as such.

SkySom (changes requested):

src/main/java/blusunrize/immersiveengineering/client/render/tile/BottlingMachineRenderer.java
MalkContent: yep! fixed

PR #4324, opened 2020-09-25
Bottling Machine Conveyor Sync

### LINKED ISSUE #4313: [MC1.15 | IE119][minor] Bottling Machine render inconsistency

MalkContent:
syncs movement of items on the Bottling Machine's conveyor belt with the belt's animation
doubles the machine's speed in the process

some code cleanup/restructuring

based on: https://github.com/BluSunrize/ImmersiveEngineering/pull/4322
because of that, also fixes: https://github.com/BluSunrize/ImmersiveEngineering/issues/4313

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/BottlingMachineTileEntity.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/BottlingMachineTileEntity.java
@@ -12,6 +12,7 @@ import blusunrize.immersiveengineering.api.DirectionalBlockPos;
 import blusunrize.immersiveengineering.api.crafting.BottlingMachineRecipe;
 import blusunrize.immersiveengineering.api.crafting.MultiblockRecipe;
 import blusunrize.immersiveengineering.api.tool.ConveyorHandler.IConveyorAttachable;
+import blusunrize.immersiveengineering.client.render.tile.BottlingMachineRenderer;
 import blusunrize.immersiveengineering.common.IEConfig;
 import blusunrize.immersiveengineering.common.IETileTypes;
 import blusunrize.immersiveengineering.common.blocks.IEBlockInterfaces.IBlockBounds;
@@ -110,8 +111,6 @@ public class BottlingMachineTileEntity extends PoweredMultiblockTileEntity<Bottl
 
 		if(isDummy()||isRSDisabled()||world.isRemote)
 			return;
-
-		tickedProcesses = 0;
 
 		int max = getMaxProcessPerTick();
 		int i = 0;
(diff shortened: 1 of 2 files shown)


BluSunrize:
I still need to check this one and test it, then I'll do a review.

BluSunrize (approved):
LGTM

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/BottlingMachineTileEntity.java
```diff
@@ -290,7 +289,8 @@ public int getProcessQueueMaxLength()
 	@Override
 	public float getMinProcessDistance(MultiblockProcess<MultiblockRecipe> process)
 	{
-		return .5f;
+		float maxTicks = BottlingProcess.getMaxProcessTick();
+		return 1f-(BottlingMachineRenderer.getTransportTime(maxTicks)+BottlingMachineRenderer.getLiftTime(maxTicks))/maxTicks;
```
malte0811: I would move these functions to the TE. I suspect that the current implementation crashes on dedicated servers, and even if it doesn't I would avoid referencing the tile render class in server-side code.

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/BottlingMachineTileEntity.java
```diff
@@ -423,7 +423,9 @@ public boolean processStep(BottlingMachineTileEntity tile)
 			if(tile.energyStorage.extractEnergy(energyExtracted, true) >= energyExtracted)
 			{
 				tile.energyStorage.extractEnergy(energyExtracted, false);
-				if(++processTick==(int)(maxProcessTick*.4375))
+				processTick++;
+				float transformationPoint = BottlingMachineRenderer.getTransportTime(maxProcessTick)+BottlingMachineRenderer.getLiftTime(maxProcessTick);
```
malte0811: Same here

malte0811 (changes requested):

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/BottlingMachineTileEntity.java
MalkContent: good point, done

PR #4327, opened 2020-09-25
[1.16.3] Simple CraftTweaker integration

kindlich:
Hello people,

a simple CrT implementation for all Recipetypes added by IE so far, except the MineralMix one, since for latter I didn't know how to properly handle the conditions and properties (though a CrT impl probably wouldn't need them).

It includes the CrT Annotation processors to create documentation entries in `./docsOut` at compilation time.
Please, when you have time to go over the changes, try to check it out into a local branch and run it, to make sure that the APs work properly in your dev environment as well. I haven't heard of anyone having issues with them so far, but then again, few know that they even exist atm...

My glorious test script can be found here:
<https://gist.github.com/kindlich/0ce8964c58c6116df328f281d3aea6da>

They are more or less the same as the examples that are in the generated docs (btw, should I put these generated files in a separate gist?).

This PR is created as draft for now since I'm going to do some more testing, and there are probably still some typos to be found in the javadocs (which will be the generated docs) and in messages...
Still, I wanted to give you the chance to go over the changes if you like ^^.

Questions:

As for the generated CrT documentation: 
I _could_ add a new step to the Jenkinsfile that will take the generated docs and uploads them to the CrT docs repository.
The again, this would increase your build pipeline time by about 15~20 seconds, and I don't think the implementation will change often enough for this to be reasonable. But your call 😛 

Do you want the MineralMix to be in there as well?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -160,6 +160,11 @@ dependencies {
     //Apparently JEI 1.16.3 can't be applied here without...
     compile fg.deobf('com.feed_the_beast.mods:ftb-gui-library:1.0.0.18')
     testImplementation('junit:junit:4.13')
+
+    compile fg.deobf(group: "com.blamejared.crafttweaker", name: "CraftTweaker-1.16.3", version: "7.0.0.45")
+    compile group: "com.blamejared.crafttweaker", name: "Crafttweaker_Annotation_Processors-1.16.3", version: "1.0.0.45"
+    annotationProcessor group: "com.blamejared.crafttweaker", name: "Crafttweaker_Annotation_Processors-1.16.3", version: "1.0.0.45"
+    annotationProcessor group: "org.reflections", name: 'reflections', version: '0.9.12'
 }
 
 def customGradle = rootProject.file('custom.gradle');
--- a/build.gradle
+++ b/build.gradle
@@ -143,16 +143,22 @@ repositories {
         //gigaherz maven for 1.16.2 mappings
         url 'https://www.dogforce-games.com/maven/'
     }
+    maven {
+        //JEI 1.16.3 tries to load 'mods:ftb-gui-library'
+        url 'https://maven.latmod.com/'
+    }
 }
 
 dependencies {
 //    minecraft "net.minecraftforge:forge:${version_minecraft}-${version_forge}"
     minecraft "net.minecraftforge:forge:1.16.3-34.0.7"
 
-    //runtimeOnly fg.deobf("mezz.jei:jei-1.16.2:7.1.0.11")
+    runtimeOnly fg.deobf("mezz.jei:jei-1.16.3:7.3.2.29")
     //runtimeOnly fg.deobf("nbtedit:NBTEdit:0.9.0")
-    compileOnly fg.deobf("mezz.jei:jei-1.16.2:7.3.2.25:api")
+    compileOnly fg.deobf("mezz.jei:jei-1.16.3:7.3.2.29:api")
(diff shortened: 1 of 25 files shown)


build.gradle
```diff
@@ -143,17 +143,28 @@ repositories {
...
     minecraft "net.minecraftforge:forge:1.16.3-34.0.7"
 
-    //runtimeOnly fg.deobf("mezz.jei:jei-1.16.2:7.1.0.11")
+    runtimeOnly fg.deobf("mezz.jei:jei-1.16.3:7.3.2.29")
     //runtimeOnly fg.deobf("nbtedit:NBTEdit:0.9.0")
-    compileOnly fg.deobf("mezz.jei:jei-1.16.2:7.3.2.25:api")
+    compileOnly fg.deobf("mezz.jei:jei-1.16.3:7.3.2.29:api")
     compile "malte0811:BlockModelSplitter:1.2.0"
+    //Apparently JEI 1.16.3 can't be applied here without...
+    compile fg.deobf('com.feed_the_beast.mods:ftb-gui-library:1.0.0.18')
```
BluSunrize, line 161: Why do we need a JEI update or a gui lib? It has worked so far without.

src/main/java/blusunrize/immersiveengineering/ImmersiveEngineering.java
```diff
@@ -96,7 +97,7 @@ public ImmersiveEngineering()
 		FMLJavaModLoadingContext.get().getModEventBus().addListener(this::wrongSignature);
 		MinecraftForge.EVENT_BUS.addListener(this::serverStarting);
 		MinecraftForge.EVENT_BUS.addListener(this::registerCommands);
-		MinecraftForge.EVENT_BUS.addListener(this::addReloadListeners);
+		MinecraftForge.EVENT_BUS.addListener(EventPriority.LOWEST, this::addReloadListeners);
```
BluSunrize, line 100: Did you verify that moving everything to lowest doesn't mess with recycling? Specifically, have you tried on servers?

BluSunrize (changes requested):
Heavy mismatch in codestyle; your bracket placement doesn't match at all.
Please follow our common style. We have an IntelliJ xml file for it in our contribution guidelines.

kindlich:
Hm... Weird, I imported that XML file, but I guess I'll do that again and run a code cleanup on the changed code

build.gradle
kindlich: In the build.gradle that is currently on that branch, JEI is only added as compile only dependency on a 1.16.2 version.
Since I'm pretty sure I'm supposed to check that the changes are in JEI, I added a 1.16.3 JEI dependency.
I first tried using the 1.16.2 version, but that gave me gradle issues, I'll try again, though.

If there's another way for me to add JEI to the runtime classpath that you prefer, I can use that instead.

src/main/java/blusunrize/immersiveengineering/ImmersiveEngineering.java
kindlich: I did some rudimentary testing on a server in deobf.
The recycling recipes were still there and usable.

Anything in particular you want looked at?

jaredlll08: From what I remember when I looked into it, I believe you need to move to 2 even listeners, one for adding recipes, and one for caching them, that would be the safest option anyway, lets mods make changes in `LOW`

kindlich:
Closing this one for now and creating another one for 1.16.4 later

PR #4334, opened 2020-09-27
Chute deploying inventory

### LINKED ISSUE #4317: [MC1.16.1 | IE108] Chutes can't be dropped into as inventories
This means that you can't drop directly into them using hoppers, dropping conveyors, item routers, etc. (as a fun fact, it DOES work with droppers though, even through the sides of a chute \^\^)

MalkContent:
gave chutes conveyor like deploying inventories
threw in a typo fix for conveyor magnet supression -> suppression

fixes: https://github.com/BluSunrize/ImmersiveEngineering/issues/4317

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/ChuteTileEntity.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/ChuteTileEntity.java
@@ -41,10 +41,14 @@ import net.minecraft.util.math.shapes.ISelectionContext;
 import net.minecraft.util.math.shapes.VoxelShape;
 import net.minecraft.util.math.shapes.VoxelShapes;
 import net.minecraft.world.World;
+import net.minecraftforge.common.capabilities.Capability;
 import net.minecraftforge.common.util.LazyOptional;
+import net.minecraftforge.items.CapabilityItemHandler;
 import net.minecraftforge.items.IItemHandler;
+import net.minecraftforge.items.IItemHandlerModifiable;
 import net.minecraftforge.items.ItemHandlerHelper;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.*;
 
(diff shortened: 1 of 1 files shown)


TeamSpen210:
Changing the API package's methods would cause a bit of an unnecessary compatibility break, would probably be better to keep the old names as an alias.

MalkContent:
fair point
I'll make it its own commit

BluSunrize:
Yeah please remove everything related to magnets from this. We can't change API stuff on the fly, that would demand an API version increase.
It gets a chunky todo of "Do this next time we bump API version" .... and then we forget about it next time we bump the API version. x3

MalkContent:
hah ^^ well it's rolled back now
the chute stuff itself never contained anti magnet stuff, besides the player-pickup prevention on deploying the item, which was already present in other parts of the chute's code

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/ChuteTileEntity.java
```diff
@@ -384,4 +388,72 @@ public boolean receiveClientEvent(int id, int arg)
...
+	@Override
+	public <T> LazyOptional<T> getCapability(@Nonnull Capability<T> cap, @Nullable Direction side)
+	{
+		if(cap==CapabilityItemHandler.ITEM_HANDLER_CAPABILITY&&side==Direction.UP)
+			return insertionCap.cast();
+		return super.getCapability(cap, side);
+	}
+
+	public static class ChuteInventoryHandler implements IItemHandlerModifiable
+	{
```
BluSunrize, line 402: This doesn't need to be Modifiable. Make it `IItemHandler` instead and ditch the `setStackInSlot` method

BluSunrize (changes requested):
Minor change request

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/ChuteTileEntity.java
MalkContent: done. This was almost a direct copy of ConveyorInventoryHandler. I don't know if that could be changed, too, or if there was a good reason for it to be IItemHandlerModifiable

malte0811 (approved):

BluSunrize (approved):
LGTM

PR #4342, opened 2020-09-30
Allow Chutes to drop into Covered Conveyors as inventories

### LINKED ISSUE #4337: [Suggestion] Make covered conveyors open their top to chutes
as title
making chutes able to directly insert into covered conveyors would also be an option, but then you couldn't transport chickens :C

MalkContent:
~~please note that I changed the IConveyorBelt API interface for that, and I don't know if that's okay:
`refactored BasicConveyor.allowCovers to .isCovered, made it public and moved it up into IConveyorBelt`~~
I had a Big Brain moment and have been proactive about it
but there's now an extra ToDo xP
resolves https://github.com/BluSunrize/ImmersiveEngineering/issues/4337


src/main/java/blusunrize/immersiveengineering/common/blocks/metal/ChuteTileEntity.java
```diff
@@ -172,7 +173,10 @@ else if(facing==Direction.EAST)
 				TileEntity inventoryTile;
 				BlockPos invPos = diagonal?getPos().offset(facing): getPos().down();
 				inventoryTile = world.getTileEntity(invPos);
-				if(!world.isRemote&&inventoryTile!=null&&!(inventoryTile instanceof IConveyorTile))
+				if(!world.isRemote&&inventoryTile!=null&&
+						(!(inventoryTile instanceof IConveyorTile)||
+								(((IConveyorTile)inventoryTile).getConveyorSubtype() instanceof BasicConveyor //TODO: change this to IConveyorTile.isCovered() when the API changes
+										&&((BasicConveyor)(((IConveyorTile)inventoryTile).getConveyorSubtype())).isCovered())))
```
malte0811: I would suggest moving this condition (starting at `inventoryTile != null`) into a separate method and structuring it along the lines of `if (a) return true; else if (b) return false; else return c`. IMO the current state is quite unreadable.

malte0811 (changes requested):
Looks good in general, just a minor code organization issue.

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/ChuteTileEntity.java
MalkContent: yea... I was debating that, too, but didn't go for it because it was single use
it was also still quite readable with the API changed and I didn't want to cause a future garbage method :|
give me a bit, I'll have it done

malte0811 (approved):

BluSunrize (approved):
LGTM

PR #3397, opened 2019-02-24
Pipes: render fix, creative scaffolding fix

MalkContent:
fixed render update for disconnecting pipes from non-pipes, made adding/removing scaffolding in creative not cost/give the scaffolding item

PR #4347, opened 2020-10-02
Pipepatch1.15

MalkContent:
This is a port of https://github.com/BluSunrize/ImmersiveEngineering/pull/3397, which didn't make it when the version jumped, I suspect.
Additionally introduces some changes because the direct port created new issues.

Fixes:
missing live connection updates
display and interaction issues with the selection bounding box
sided fluidhandler being visible when it probably shouldn't be

Adds:
selection bounding box giving interactable preview of new connections when holding a hammer

This one was much more than I initially thought it would be and had lots of new stuff in it. I think I was very diligent with making sure I knew what I did, before I did it, and I tested a lot to make sure that everything still works as it should. But beware.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/FluidPipeTileEntity.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/FluidPipeTileEntity.java
@@ -20,7 +20,6 @@ import blusunrize.immersiveengineering.common.blocks.IEBaseBlock.IELadderBlock;
 import blusunrize.immersiveengineering.common.blocks.IEBaseTileEntity;
 import blusunrize.immersiveengineering.common.blocks.IEBlockInterfaces.*;
 import blusunrize.immersiveengineering.common.blocks.IEBlocks.WoodenDecoration;
-import blusunrize.immersiveengineering.common.items.IEItems.Tools;
 import blusunrize.immersiveengineering.common.util.CapabilityReference;
 import blusunrize.immersiveengineering.common.util.Utils;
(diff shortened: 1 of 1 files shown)


src/main/java/blusunrize/immersiveengineering/common/blocks/metal/FluidPipeTileEntity.java
```diff
@@ -579,19 +624,24 @@ public boolean receiveClientEvent(int id, int arg)
...
 
 	@Override
 	public VoxelShape getSelectionShape(@Nullable ISelectionContext ctx)
 	{
-		return SHAPES.get(new BoundingBoxKey(false, this));
+		boolean hasHammer;
+		if(ctx!=null&&ctx.hasItem(Tools.hammer))
+			hasHammer = true;
+		else
+			hasHammer = false;
```
BluSunrize: I would prefer this to use `Utils.isHammer(ItemStack)`, however you will need to check the ctx to be EntitySelectionContext and then retrieve the MAIN_HAND item of the entity involved.

That is more complex, but our utility method should be used whereever possible, so that if our logic for detecting a hammer changes (currently using tool classes, may allow tags in the future), we only have to change it in one place.

BluSunrize (changes requested):
Minor change request

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/FluidPipeTileEntity.java
MalkContent: neat. Done!

PR #4351, opened 2020-10-04
Excavator mirrored bucketwheel flicker fix

MalkContent:
mirrors bucketwheel upon mirrored excavator creation, removing the flicker of it jumping from non-mirrored to mirrored
this also might make reorienting and mirroring during the excavators tick unnecessary (didn't get called anymore in debugging for the cases I worked through, even stuff like building a mirrored excavator around a non-mirrored bucketwheel and the other way around), but I'll let you make that call


BluSunrize:
But Malte already addressed the jumpyness in this commit, didn't he?

https://github.com/BluSunrize/ImmersiveEngineering/commit/70ad591d787d1b954b5682f1c87bbc18613f149b

MalkContent:
Malte's seems like a different thing, addressing animation. This here is a super minor thing I noticed during the last excavator PR and it kept being something in the back of my head.
When you create a mirrored excavator and let it create a wheel at the same time, it will create a non-mirrored wheel and immediately snap it to mirrored in the excavator-tile's tick, which I called "flicker" here.

BluSunrize (approved):
LGTM.
@malte0811 any objections?

malte0811 (approved):

PR #4371, opened 2020-10-12
[1.16.3] Anti IndexOutOfBounds measure

TwistedGate:
`tile.targetList.remove(rem)` is to blame


BluSunrize:
I'll cherry-pick this into 1.15 at a later point instead, because merging and then picking will be messy.

BluSunrize:
Picked!

PR #4387, opened 2020-10-17
Reduce size of recipes in PacketBuffer's

malte0811:


BluSunrize:
I understand that this is a quick fix to the recipes added by IE, but shouldn't issues with vanilla shaped recipes be handled in Forge instead?

malte0811:
No, Forge servers need to be compatible with vanilla clients and vice versa. With this changed format that isn't (easily?) possible.

BluSunrize:
Ah! I was unaware/ignorant of the fact that Forge and Vanilla required compatability.
That said, Compressed Recipes could still be a part of Forge's framework, much like ShapedOre and ShapelessOre recipes were in past iterations, no?
I'm not looking to dismantle this PR, I'm all for reducing the packet size, I'm just thinking that it would be good to standardize this across the modded community!

BluSunrize:
@malte0811 Are you still looking to have this implemented?

malte0811:
Probably some parts of it, but not everything. Some things are optimized much better in RecipeBuffers than they can be here, but other things I can't really influence from outside IE. I'll remove the parts that I don't want any more when I have some time.

PR #4393, opened 2020-10-18
Sand Recipe Updates

MalkContent:
- allow Cactus and Sugar Cane to use Red Sand as soil in the Cloche
- make Red Sandstones in the Crusher return Red Sand instead of regular one
- changed ClocheRecipeBuilder in the API to also allow Tags as parameter for addSoil


BluSunrize (approved):
I didn't realize that Forge hadn't tagged sandstone in the same way they tagged sand; but this looks like the correct way to do it.

LGTM.

Issue #4389, opened 2020-10-17
[1.16.3] FireworkBullet in Turret->Crash


mpmxyz:
Somebody mentioned this issue in a sidenote  (#4348) but didn't have a crash log:

[Here it is!](https://gist.githubusercontent.com/mpmxyz/de9e3c75d01746bb1fbd718cb088b799/raw/3057f1058dbae7446c7c6e106d826997bd911aeb/gistfile1.txt)

### How to reproduce
Enable a gun turret with a firework bullet - the game crashes when it fires. (implies having energy and a target)

### Cause of Crash
[TurretGunTileEntity.activate](https://github.com/BluSunrize/ImmersiveEngineering/blob/03151dccc5e500ef3a2cbfb72f1756d6cbb9c7f3/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/TurretGunTileEntity.java#L106) calls [FireworkBullet.getProjectile](https://github.com/BluSunrize/ImmersiveEngineering/blob/03151dccc5e500ef3a2cbfb72f1756d6cbb9c7f3/src/main/java/blusunrize/immersiveengineering/common/items/BulletItem.java#L436) with parameter 'shooter' == null
The crash is caused by [some code](https://github.com/BluSunrize/ImmersiveEngineering/blob/03151dccc5e500ef3a2cbfb72f1756d6cbb9c7f3/src/main/java/blusunrize/immersiveengineering/common/items/BulletItem.java#L440) not expecting a null argument.

### Versions & Modlist
Forge: 34.1.0
IE: 1.16.3-4.1.0-124

PR #4394, opened 2020-10-18
[1.6.3] Fixed: crashes when using Wolfpack/Firework Cartridges in a Gun Turret

mpmxyz:
I filed #4389 but figured that if I can find the root cause (shooter==null) I should be able to fix it, too.
When testing I found another bug with a similar cause that I also fixed.
I tested the revolver and the gun turret with all kinds of cartridges and there were no crashes after this fix.
Other things I found: Fireworks are OP against flocks of slimes; slow moving/gravity affected projectiles are not aimed properly.

PS: I didn't have the testing/development capacity + git-knowledge to easily adjust the code for lower Minecraft versions. So one can consider this a bit incomplete.


BluSunrize:
I already have a fix for #4389 in the pipeline, it's just not pushed. I wrote it almost exactly 2 hours ago. I'll look into the wolfpack issue though, thanks for bringing it up.

PR #4433, opened 2020-11-05
Move (most) ATs and all coremods to Mixins

malte0811:
The first two commits are mostly "mechanical" (replacing the ATs), the third one is probably the most interesting one.


SkySom (approved):

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -1,59 +1,4 @@
...
-# Used for village generation
-public net.minecraft.world.gen.feature.jigsaw.LegacySingleJigsawPiece <init>(Lcom/mojang/datafixers/util/Either;Ljava/util/function/Supplier;Lnet/minecraft/world/gen/feature/jigsaw/JigsawPattern$PlacementBehaviour;)V
-# Used to mark Fabulous as unsupported when stencil is enabled
-public net.minecraft.client.renderer.GPUWarning field_241688_c_ # warnings map
-public net.minecraft.world.gen.feature.template.Template$Palette <init>(Ljava/util/List;)V # Template.Palette constructor
-public net.minecraft.world.gen.feature.template.Template field_186272_c # size
+# Used in ModelPowerPack, remove final modifier
+public-f net.minecraft.client.renderer.model.Model func_228282_a_(Lnet/minecraft/util/ResourceLocation;)Lnet/minecraft/client/renderer/RenderType; #getRenderType
+# Used in TurnAndCopyRecipe, make protected to allow overrides
+protected net.minecraft.item.crafting.ShapedRecipe func_77573_a(Lnet/minecraft/inventory/CraftingInventory;IIZ)Z #checkMatch
```
BluSunrize, line 2: Is there a reason we are keeping these?

BluSunrize (approved):
LGTM

src/main/resources/META-INF/accesstransformer.cfg
malte0811: Yes, we override both of these, which (as far as I can tell) isn't possible with Mixins. The first one is final, so we can't override it without removing that modifier; the second one is private, so we need to make it (at least) protected.

I just noticed that we can actually get rid of the first one cleanly as that method just applies a `Function` stored in a field (will push that in a minute), removing the second one would require copying large parts of `ShapedRecipe`, so I'm keeping it for now.

malte0811: Actually, thought: we don't really need the override for anything but the "turn" functionality. It might be an idea to move that functionality to datagen (i.e. generate one "unturned" and one "rotate by x degrees" recipe), then we could probably get rid of ATs completely.

BluSunrize: That would mean listing multiple recipes in JEI. I don't want that.

PR #4457, opened 2020-11-22
Improved Belljar performance

tth05:
Not sure if you still accept 1.12 PRs, but I'll try anyway...

I improved the redstone checking code and reduced the client side `IPlantHandler#isValid` checks. These checks could be optimized on the server side as well, but the custom inventory used here is just a list without some onChanged method where the plant handler could be invalidated.


BluSunrize:
1.12 is no longer being maintained.

PR #4459, opened 2020-11-23
Use VoxelShapes to determine if there's space for dropping conveyors to drop

### LINKED ISSUE #4454: 1.16.4 -- Change of behavior in interaction between Crushers and Dropper Belts
### Description of the issue:
In IE for 0.12-89, on MC 1.12.2, a Dropper Belt placed straight on a Crusher's bin will place items in the Crusher's inventory, who will start working immediately. Edit: It will not insert items the Crusher cant process.

In IE 1.16.4-4.1.1-128, on MC 1.16.4 a Dropper Belt placed straight on a Crusher's bin will not drop any items into it.

Included, a1.16.4 of how the Dropper Belt was placed in both versions.
unknown

### Crashlog:
There is no Crashlog

### Versions & Modlist
Minecraft 1.16.4; Forge 35.0.18; IE 1.16.4-4.1.1-128

TeamSpen210:
This way, the existing special cases for air, trapdoors and other conveyors just works naturally, and dropping into things like crushers also works (fixes #4454). Since the combine function is somewhat expensive, I added a cache for that.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/conveyors/DropConveyor.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/conveyors/DropConveyor.java
@@ -36,7 +36,8 @@ import static blusunrize.immersiveengineering.ImmersiveEngineering.MODID;
 public class DropConveyor extends BasicConveyor
 {
 	public static final ResourceLocation NAME = new ResourceLocation(MODID, "dropper");
-	private static final VoxelShape REQ_SPACE = VoxelShapes.create(0.25, 0.75, 0.25, 0.75, 1.0, 0.75);
+	/// Items will be spawned when this space is empty in the block below:
+	private static final VoxelShape REQUIRED_SPACE = VoxelShapes.create(0.25, 0.75, 0.25, 0.75, 1.0, 0.75);
 	private VoxelShape cachedDownShape = VoxelShapes.empty();
 	private boolean cachedOpenBelow = true;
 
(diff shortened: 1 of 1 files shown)


src/main/java/blusunrize/immersiveengineering/common/blocks/metal/conveyors/DropConveyor.java
```diff
@@ -84,14 +89,17 @@ else if(contact&&isEmptySpace(getTile().getWorld(), posDown, inventoryTile))
...
-		if(state.getBlock() instanceof TrapDoorBlock)
-			return state.get(TrapDoorBlock.OPEN);
-		return false;
+		VoxelShape shape = state.getCollisionShape(world, pos);
+		// Combining voxelshapes is a little expensive, so only calculate
+		// when the voxelshape changes. Identity compare is sufficent since they're
+		// usually precomputed.
+		if (shape != cachedDownShape)
+		{
+			cachedOpenBelow = VoxelShapes.compare(REQ_SPACE, shape, IBooleanFunction.AND);
```
malte0811: I think you're missing a `!` here, this gives `false` for `shape = VoxelShapes.empty()`.

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/conveyors/DropConveyor.java
```diff
@@ -34,6 +36,9 @@
 public class DropConveyor extends BasicConveyor
 {
 	public static final ResourceLocation NAME = new ResourceLocation(MODID, "dropper");
+	private static final VoxelShape REQ_SPACE = VoxelShapes.create(0.25, 0.75, 0.25, 0.75, 1.0, 0.75);
```
malte0811: I would prefer a slightly longer name here (probably `REQUIRED_SPACE`), but it's not that important.

malte0811 (changes requested):

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/conveyors/DropConveyor.java
TeamSpen210: Indeed.

malte0811 (approved):

BluSunrize:
Approved. Thanks for the contribution =)

BluSunrize:
...I probably should have rebased this to 1.16.4
fuck.
I'll fix that locally.

PR #4464, opened 2020-11-28
Make the Engineer's Toolbox able to accept items by tags

ConductiveFoam:


TeamSpen210:
It might be easier/better to convert the toolbox to entirely use tags - just have `immersiveengineering:toolbox_tools` which then includes the default items.

Issue #4467, opened 2020-11-29
[Suggestion] Allow copying Blueprints for use by other players


Algester:
Since these are sometimes a pain to acquire in the newer versions of minecraft cause can't decide when RNG is in your favor I'm suggesting that the BPs could be copied by making some sort of carbon paper and or blank blueprint for the players to use and copy such rare blueprints

TeamSpen210:
It'd also be useful if you wanted to build multiple automated workbenches, or other stuff like that.

PR #4469, opened 2020-12-01
Implement #4467: Add Blueprint copying recipe

TeamSpen210:
This implements #4467 by adding a recipe for duplicating any blueprint, with 3 paper and 3 dye plus the existing blueprint. The recipe layout is fully configurable in the JSON, if users want to change that. I ATed some helpers from `ShapedRecipe` for parsing the JSON, should I rebase onto #4433 and use Mixins for that instead?


BluSunrize:
We don't do access transformer anymore, that's why Malte has a PR for moving everything to Mixins.

Also, I have plans (tm) for blueprints, so I want to handle this differently in the future. I appreciate the contribution, but it doesn't match what I had in mind.

PR #4474, opened 2020-12-02
Fix fluid ingredients

### LINKED ISSUE #4449: Mekanism's tank is not treated as a "bucket" when completing the crafting of treated wood
### Description of the issue:
mekanism's tank is not being treated as a returned empty bucket on making treated wood with Creosote Oil, I'll relay this info on Mekanism's git if someone thinks this is on mekanism's end

### Versions & Modlist
IE: 1.16.3-4.1.1-127
Forge: 34.1.42
Mekanism: 1.16.4-10.0.17.444

malte0811:
Closes #4449

I do not recommend looking at the individual commits, just looking at the final result will be less confusing. Also includes a fix for failing datagen runs (broken by #4433), I still need to look into making that kind of failure noticeable in the GH actions.  
Tested with buckets, jerrycans and Mekanism tanks, results are all as expected. Also tested with the IE assembler.

cc @pupnewfster


src/main/java/blusunrize/immersiveengineering/api/crafting/FluidTagInput.java
```diff
@@ -170,4 +173,22 @@ public void write(PacketBuffer out)
...
+
+	public boolean extractFrom(IFluidHandler handler)
+	{
+		int remaining = amount;
+		for(int tank = 0; tank < handler.getTanks(); tank++)
+		{
+			FluidStack inTank = handler.getFluidInTank(tank);
+			if(testIgnoringAmount(inTank))
+			{
+				int extractAmount = Math.min(inTank.getAmount(), remaining);
```
pupnewfster: Given how the fluid handler API is I am not sure if this line makes sense, because the drain implementation isn't requested from a specific tank but left up to the fluid handler to decide how to implement, there should be no harm in requesting to extract remaining, even if the specific tank you are looking at type wise contains less contents. If more can't be extracted the drain method should not return more than was actually able to be extracted.

src/main/java/blusunrize/immersiveengineering/common/crafting/fluidaware/AbstractFluidAwareRecipe.java
```diff
@@ -0,0 +1,129 @@
...
+			{
+				final int invIndex = getInventoryIndex(inv, x, y);
+				final int ingrIndex = offset.getListIndex(x, y);
+				if(ingrIndex >= 0)
+				{
+					Ingredient ingr = getIngredients().get(ingrIndex);
+					final ItemStack item = inv.getStackInSlot(invIndex);
+					ItemStack result = ItemStack.EMPTY;
+					if(ingr instanceof IngredientFluidStack)
+						result = ((IngredientFluidStack)ingr).getExtractedStack(item).copy();
```
pupnewfster: I am not positive but I believe it would make sense to copy the item here (or in getExtractedStack before extracting), just so that if a crafting implementer queries what remaining items are for a recipe there are no side effects of the container being drained just because of the getter being called to say see if an automated crafter has room to also store the remaining stacks.

pupnewfster (commented):
Only skimmed through your refactoring and mostly ignored all the MatchLocation stuff etc, but added a couple comments of things you may want to potentially change.

src/main/java/blusunrize/immersiveengineering/common/crafting/fluidaware/IngredientFluidStack.java
```diff
@@ -73,8 +77,12 @@ public boolean test(@Nullable ItemStack stack)
 	{
 		if(stack==null)
 			return false;
-		else
-			return this.fluidTagInput.test(FluidUtil.getFluidContained(stack).orElse(FluidStack.EMPTY));
+		// Copy the stack here, that allows us to be (more) sure that we can actually extract the required amount (and
+		// won't run into issues where extracting from one tank affects the other one)
+		Optional<IFluidHandlerItem> handler = FluidUtil.getFluidHandler(stack).resolve();
+		if(!handler.isPresent())
```
SkySom: wouldn't this just been simplier as 
```
return handler.isPresent() && ...
```

SkySom (approved):

PR #4491, opened 2020-12-10
fix crash with equipped voltmeter while flying

gyroplast:


malte0811:
I am very happy about finally having a reproducible test case of this issue (you have been the first to mention voltmeters in the context of this crash), but this is unfortunately not a good fix for the problem. With just this change, client-side bugs, client-side mods changing behavior (e.g. Optifine) and malicious client-side code (hopefully unlikely, but still) can still cause server-side crashes, which is generally a very bad idea. I've implemented a server-side fix, I'll push that in a few minutes.

PR #4493, opened 2020-12-10
Fix jerrycans disappearing when filling empty buckets, closes #4482

### LINKED ISSUE #4482: [1.16.4] Jerrycans disappear when filling buckets
### Description of the issue:
When filling empty buckets from a jerrycan, the jerrycan disappears after one bucket gets filled. This happens regardless of the can's fill state.

### Versions & Modlist
IE 1.16.4-4.1.1-128
Forge 35.1.6 and 35.1.0, latest and recommended respectively at time of writing

ConductiveFoam:
I *hope* this is an acceptable fix; it seems to me that the person writing it before just got the operator inverted.


malte0811 (changes requested):
That fix is definitely not correct. `getFluidHandler` returns a `LazyOptional`, which (both by convention and in practice) is non-null. So comparing that value to null is meaningless. The proper fix would probably be something involving `LazyOptional.isPresent()`.

Make sure to also test your fix with filling a jerrycan from another jerrycan, that case currently seems to be broken as well.

ConductiveFoam:
Came up with a different implementation for `getRemainingItems`: Keep only the first recognized Jerrycan, since that's the one that `getCraftingResult` uses as a source of fluid. All other items get discarded as before; this is also is why 9515e4d restricts recipes to two items, before one could lose items by filling more of the slots.
Now Jerrycan-Jerrycan filling works too

Hope this is better!

src/main/java/blusunrize/immersiveengineering/common/crafting/JerrycanRefillRecipe.java
```diff
@@ -87,6 +87,16 @@ else if(ret[1].isEmpty()&&FluidUtil.getFluidHandler(stackInSlot).isPresent())
 		return ret;
 	}
 
+	private int countOccupiedSlots(IInventory inv) {
+		int c = 0;
+		for(int i = 0; i < inv.getSizeInventory(); i++)
+		{
```
malte0811: Formatting (the braces should be removed)

src/main/java/blusunrize/immersiveengineering/common/crafting/JerrycanRefillRecipe.java
```diff
@@ -87,6 +87,16 @@ else if(ret[1].isEmpty()&&FluidUtil.getFluidHandler(stackInSlot).isPresent())
 		return ret;
 	}
 
+	private int countOccupiedSlots(IInventory inv) {
```
malte0811: Formatting (brace should be on the next line)

src/main/java/blusunrize/immersiveengineering/common/crafting/JerrycanRefillRecipe.java
```diff
@@ -98,12 +108,18 @@ public boolean canFit(int width, int height)
...
-			if(FluidUtil.getFluidHandler(inv.getStackInSlot(i))!=null)
+			ItemStack stackInSlot = inv.getStackInSlot(i);
+			if(!stackInSlot.isEmpty())
 			{
+				if(Misc.jerrycan.equals(stackInSlot.getItem())&&!foundJerrycan)
+				{
+					foundJerrycan = true;
+					continue;
+				}
 				remaining.set(i, ItemStack.EMPTY);
```
malte0811: IMO something like
```java
if (…)
    foundJerrycan = true;
else
    remaining.set(i, ItemStack.EMPTY);
```
would be nicer here.

malte0811 (commented):
Looks good now (aside from some minor formatting problems), and seems to fix the problem it's intended to fix. There's another problem with this recipe (filling by non-1000-mB-amounts is handled by essentially using a side-channel, which seems to be broken), but that's not related to the problem described in #4482. I'll look into that problem later, unless you want to include that in this PR?

ConductiveFoam:
Glad it looks good!
Fixed the formatting problems.

I don't get what other problem you are referring to, filling odd amounts works for me. The tooltip in the crafting grid sometimes is off, the items themselves get processed fine however. Can you elaborate? I'm happy to help out some more

BluSunrize:
...shit I merged this into the wrong branch

PR #4494, opened 2020-12-10
Cap maximum furnace burn time of creosote containing items to one bucket's worth

ConductiveFoam:
When a FurnaceFuelBurnTimeEvent resolves its item's contained fluid amount is reduced by up to one bucket so the calculation for the creosote burn time needs to reflect that.


malte0811:
The PR itself is fine, but I'm not sure about this feature in general. It has the same issues as the fluid recipes had previously, i.e. it makes an assumption about fluid containers that isn't necessarily true (and isn't true for Mek's tanks). In this case we do not have any way of replacing the container item, so IMO the feature should be removed/limited to IE containers (jerrycan and buckets). @BluSunrize will need to make the final decision on that though.

BluSunrize:
I'd rather remove it, in that case. I'm tired of accounting for various edgecases, and other mods that don't use container items in sensible ways.

PR #4499, opened 2020-12-12
Make sawblades and drillheads explicitly non-enchantable (fixes #4480)

### LINKED ISSUE #4480: Unbreaking should not work on Drillheads and Sawblades
Title says it all. Fuck knows why Mojang allows applying that to *anything* in an anvil, but it's definitely not intended for IE, so we should look into preventing it.

ConductiveFoam:


BluSunrize:
Thank you =)

PR #4505, opened 2020-12-17
Minor fixes for MineralVeins with null mixes

SkySom:
Properly not 100% fixed, but should reduce heavily.


BluSunrize (approved):
Looks good to me, but I'd like Malte's input too

malte0811 (approved):

PR #4563, opened 2021-01-10
[1.16.4] CraftTweaker integration

kindlich:
Basic integration for the different Recipetypes added by IE
Also adds the the crafttweaker command `/ct ieBlueprintCategories` and integration with the `/ct examples` command.

I have no gist with a link to example script files here, as they are part of the PR (check the data directory)

Still a draft PR, since I only tested on SinglePlayer so far, so I'd feel uncomfortable with having it merged yet. Due to university and work I don't know when I can get to test the scripts thoroughly on a server.

The code should be ready for review though ^^
Some of the `@docparam` arguments need updating

Also, I have no clue why the GH action fails with the CrT dependency, running the build locally works fine


jaredlll08:
@SkySom this is ready for review (just mark it as not a draft)

src/main/java/blusunrize/immersiveengineering/common/util/compat/crafttweaker/CrTIngredientUtil.java
```diff
@@ -0,0 +1,90 @@
...
+
+	private CrTIngredientUtil()
+	{
+	}
+
+	/**
+	 * So far, only IItemStack supports setting an amount.
+	 * So we can at least check for that, I guess?
+	 */
+	public static IngredientWithSize getIngredientWithSize(IIngredient crafttweakerIngredient)
```
SkySom, line 35: Would it make sense to have count somehow be applied within CraftTweaker to IIngredients? I also use count on Tags and such in Transport, and definitely see it as a useful addition (And I think it is something that is useful? Only having ItemStacks be able to have count for the integration with IE seems slightly less useful?)

SkySom (approved):
May want to look into why your PR is grabbing additional commits on it's own. Note the comment on count with IIngredients.

BluSunrize:
Closing this in favor of #4577

PR #4572, opened 2021-01-16
update Japanese translation

karakufire:


koh-gh:
I suggest some improvements.

- Hoes are translated as 'クワ' in Vanilla. Therefore, it is preferable that steel hoe be translated in the same way.(Use katakana translation)
- Liquid creosote is translated as 'クレオソートオイル', while bucket creosote is translated as 'クレオソート油'. They need to be the same.

And, there are some that haven't been translated yet. Are there any plans for additional translations?

karakufire:
Thank you for reviewing.
These 2 points you raised are correct and I'll change them.
And untranslated entries, 
* shaders' names, "Oresome": Because I couldn't find suitable translation for them, I keep as is in respect for original.
* ie.manual.entry.mineralsDimInvalid, ie.manual.entry.mineral_main: Because I could not check where them would be used.

Untranslated shader's description(excepts "waaagh") is simply I forgot to translate.

BluSunrize:
Is this ready to be merged?

koh-gh:
I have no problem.

karakufire:
I'm OK

BluSunrize:
Thank you =)

PR #4577, opened 2021-01-17
Feature/crafttweaker integration

jaredlll08:
This supersedes #4563 and targets the 1.16.5 branch.

This is marked as a draft while we discuss @SkySom's comment on $4563
> Would it make sense to have count somehow be applied within CraftTweaker to IIngredients? I also use count on Tags and such in Transport, and definitely see it as a useful addition (And I think it is something that is useful? Only having ItemStacks be able to have count for the integration with IE seems slightly less useful?)


SkySom (commented):
Looks good to me. Will pop to approval once the count stuff is fixed it.

src/main/java/blusunrize/immersiveengineering/common/util/compat/crafttweaker/CrTIngredientUtil.java
```diff
@@ -0,0 +1,90 @@
...
+	 * So far, only IItemStack supports setting an amount.
+	 * So we can at least check for that, I guess?
+	 */
+	public static IngredientWithSize getIngredientWithSize(IIngredient crafttweakerIngredient)
+	{
+		final Ingredient basePredicate = crafttweakerIngredient.asVanillaIngredient();
+		if(crafttweakerIngredient instanceof IItemStack)
+		{
+			return IngredientWithSize.of(((IItemStack)crafttweakerIngredient).getInternal());
+		}
```
malte0811: Formatting (braces should be omitted for single-statement blocks). This is consistent in the entire diff, so I won't comment on other cases.

malte0811 (commented):
Other than the formatting issue the PR looks good to me.

BluSunrize (approved):
Ideally the copyright notices should say 2021 on them, given that that's the their commit date as well. But beyond that, LGTM

src/main/java/blusunrize/immersiveengineering/common/util/compat/crafttweaker/CrTIngredientUtil.java
jaredlll08: Applied the codestyle, so this should all be good now ^^

jaredlll08:
Update (and added where it was missing) the license header, this should be good to go!

SkySom (approved):

PR #4651, opened 2021-02-13
translate rus 1.7.10

GemMaxweLlX:


malte0811:
1.7.10 will not get any new releases, it has been unsupported since early 2016.

GemMaxweLlX:
> 1.7.10 will not get any new releases, it has been unsupported since early 2016.

sorry for the commit I'm only playing on 1.7.10

PR #4688, opened 2021-03-04
Adds BiR and ER compat for ThermoElectric Generator

OneLemonyBoi:


BluSunrize (changes requested):
Mod compat belongs in a proper compat module, not the IEContent class.

PR #4706, opened 2021-03-11
Move API to separate source set

malte0811:
It's probably a good idea to look at this locally with something along the lines of
```
git diff  --ignore-all-space origin/1.16.5 HEAD
```
since quite a few files still had windows line ends. Even that doesn't have perfect diff, since some files changed enough for git to consider them different files rather than moved and modified.


src/api/java/blusunrize/immersiveengineering/api/tool/ConveyorHandler.java
```diff
@@ -70,8 +63,9 @@
 	public static final Map<Class<? extends IConveyorBelt>, ResourceLocation> reverseClassRegistry = new LinkedHashMap<>();
 	public static final Set<BiConsumer<Entity, IConveyorTile>> magnetSupressionFunctions = new HashSet<>();
 	public static final Set<BiConsumer<Entity, IConveyorTile>> magnetSupressionReverse = new HashSet<>();
```
BluSunrize: These should be changed to "Suppression". I forgot the second P ages ago, and never fixed it, because API breaking =P

src/api/java/blusunrize/immersiveengineering/api/wires/utils/WireUtils.java
```diff
@@ -53,7 +55,7 @@ public static Connection raytraceWires(World world, Vector3d start, Vector3d end
 		WireCollisionData collisionData = global.getCollisionData();
 		AtomicReference<Connection> ret = new AtomicReference<>();
 		AtomicDouble minDistSq = new AtomicDouble(Double.POSITIVE_INFINITY);
-		Utils.rayTrace(start, end, world, (pos) ->
+		RAYTRACE.getValue().rayTrace(start, end, world, (pos) ->
```
BluSunrize: Couldn't we just have moved rayTrace to an API utility class?

src/api/java/blusunrize/immersiveengineering/api/IEProperties.java
```diff
@@ -196,8 +200,6 @@ public int hashCode()
 		public static final ModelProperty<Map<String, String>> TEXTURE_REMAP = new ModelProperty<>();
 		public static final ModelProperty<ConnectionModelData> CONNECTIONS = new ModelProperty<>();
 		public static final ModelProperty<MineralMix[]> MINERAL = new ModelProperty<>();
-		public static final ModelProperty<IConveyorBelt> CONVEYOR = new ModelProperty<>();
-		public static final ModelProperty<FeedthroughData> FEEDTHROUGH = new ModelProperty<>();
 		public static final ModelProperty<Map<Direction, IOSideConfig>> SIDECONFIG = new ModelProperty<>();
 		public static final ModelProperty<BlockPos> SUBMODEL_OFFSET = new ModelProperty<>();
 		//TODO remove?
```
BluSunrize: Well? Do we still need this? Either we remove the property or the TODO. Now would be the chance.

src/api/java/blusunrize/immersiveengineering/api/wires/utils/WireUtils.java
```diff
@@ -251,4 +233,21 @@ public static void moveConnectionEnd(Connection conn, ConnectionPoint currEnd, C
 		globalNet.removeConnection(conn);
 		globalNet.addConnection(new Connection(conn.type, fixedPos, newEnd));
 	}
+
+	public interface RaytraceCallback
+	{
+		HashSet<BlockPos> rayTrace(Vector3d start, Vector3d end, World world, Consumer<BlockPos> out);
+	}
+
+	//TODO dedup with Utils#…
```
BluSunrize: Are you going to resolve this TODO as part of this PR, or in a later step?

BluSunrize (approved):
Overall this looks good to me.

I'm not too big of the fan of the frequent use of SetRestrictedField. I understand what it does and why it's necessary, I just don't like it aesthetically. It also feels like at least some of those utility methods could just have been moved to the API itself.

I only spotted one typo that I had rembered wanting to fix. I hope that's all of them.

src/api/java/blusunrize/immersiveengineering/api/IEProperties.java
malte0811: I guess making it non-API would be an option. The problem is that on the one hand passing the complete TE defeats the point of model properties, is a threading nightmare and makes caching hard; while on the other hand a lot of code currently uses this functionality in ways that aren't trivial to change.

src/api/java/blusunrize/immersiveengineering/api/wires/utils/WireUtils.java
malte0811: I'll check it later today, that's definitely for this PR. I'm pretty sure I've resolved the TODO at some point, either I forgot to remove it after resolving it or that change got lost somewhere.

PR #4781, opened 2021-04-11
CC compat

malte0811:
Don't worry about the 4,594 deletions, that's mostly the 1.12 compat code for OC and CT that I removed from `oldjava` since we have new implementations of both now.

As I said before I'm roughly following pup's idea of having one generic "can be called by a computer" concept which should be easy to use for other computer mods in the future. It does involve some weird `MethodHandle`-code, but that part can be treated as a black box most of the time.


src/main/java/blusunrize/immersiveengineering/common/util/compat/computers/cctweaked/ComputerCraftCompatModule.java
```diff
@@ -0,0 +1,132 @@
+/*
+ * BluSunrize
+ * Copyright (c) 2020
+ *
```
BluSunrize, line 4: Should these files not use a 2021 copyright statement?

src/main/java/blusunrize/immersiveengineering/common/util/compat/computers/generic/IndexArgument.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicates that the annotated parameter is an index which is 1-based in Lua and 0-based in Java. The index will be
+ * automatically shifted when calling the method.
+ */
```
BluSunrize, line 19: I continue hate Lua for this. But I'm glad you have a solution for it.

src/main/java/blusunrize/immersiveengineering/common/util/compat/computers/generic/impl/PoweredMBCallbacks.java
```diff
@@ -0,0 +1,16 @@
...
+public class PoweredMBCallbacks extends Callback<PoweredMultiblockTileEntity<?, ?>>
+{
+	public static final PoweredMBCallbacks INSTANCE = new PoweredMBCallbacks();
+
+	public PoweredMBCallbacks()
+	{
+		addAdditional(EnergyCallbacks.INSTANCE);
+		addAdditional(MultiblockCallbacks.INSTANCE);
+	}
+	//TODO general access to recipes?
```
BluSunrize, line 23: Hmm. Unsure on that. Seems very potent.

BluSunrize (approved):
LGTM.
No license statements on any of the new files. Is that intentional?

src/main/java/blusunrize/immersiveengineering/common/util/compat/computers/cctweaked/ComputerCraftCompatModule.java
malte0811: This one isn't new, I just moved it and then modified it enough for git to call it new. I can change the copyright year on it if you want. The other files not having any copyright statement is definitely not intentional, I must have lost some configuration in IntelliJ at some point.

src/main/java/blusunrize/immersiveengineering/common/util/compat/computers/generic/impl/PoweredMBCallbacks.java
malte0811: I was talking about access to the current recipe, so you could e.g. make a display "currently converting plant oil and ethanol to biodiesel at a ratio of 2:2:4". That seems reasonable from a potency perspective to me. But I've essentially put that on the list of "implement this once I find an elegant approach for it".

BluSunrize: Ah, I see

src/main/java/blusunrize/immersiveengineering/common/util/compat/computers/cctweaked/ComputerCraftCompatModule.java
BluSunrize: Right, it showed as new file, I didn't realize it was one of the moved ones.

PR #4810, opened 2021-05-06
Feature: Redstone Logic

BluSunrize:
Adds the circuit table, logic circuits and a logic unit to run them.

Took me a while to get this finished, now it just has to pass review :'D


src/api/java/blusunrize/immersiveengineering/api/tool/LogicCircuitHandler.java
```diff
@@ -0,0 +1,182 @@
...
+		NOR(2, args -> !(args[0]|args[1]), 4),
+		NAND(2, args -> !(args[0]&args[1]), 1),
+		XNOR(2, args -> args[0]==args[1], 5);
+
+		private final int argumentCount;
+		private final Predicate<boolean[]> operator;
+		private final int complexity;
+
+		public static final int TOTAL_MAX_INPUTS = Arrays.stream(values())
+				.map(LogicCircuitOperator::getArgumentCount).max(Integer::compareTo).orElse(1);
```
malte0811: I feel like a few extra line breaks might be nice here (I usually split at every call after `stream()`). And maybe `mapToInt` instead of `map`? Then you don't need an explicit comparator for `max`.

src/main/java/blusunrize/immersiveengineering/client/gui/elements/GuiButtonLogicCircuitRegister.java
```diff
@@ -0,0 +1,88 @@
...
+	@Override
+	public boolean keyPressed(int keyCode, int scanCode, int modifiers)
+	{
+		if(keyCode >= 48&&keyCode <= 55) // digits 0-8
+		{
+			this.setStateByInt(keyCode-48+16); // keycode to zero, plus 16 colors
+			this.onPress.onPress(this);
+			return true;
+		}
+		else if(keyCode >= 65&&keyCode <= 90) // A-Z
```
malte0811: I'd use the constants in `GLFW`, or possibly `'0'`/`'8'`/etc

src/main/java/blusunrize/immersiveengineering/common/network/MessageContainerUpdate.java
```diff
@@ -0,0 +1,58 @@
...
+	{
+		Context ctx = context.get();
+		ServerPlayerEntity player = ctx.getSender();
+		assert player!=null;
+		ctx.enqueueWork(() -> {
+			player.markPlayerActive();
+			if(player.openContainer.windowId==windowId&&player.openContainer instanceof IEBaseContainer)
+			{
+				((IEBaseContainer<?>)player.openContainer).receiveMessageFromScreen(nbt);
+			}
```
malte0811: Braces on single-line `if`

src/main/java/blusunrize/immersiveengineering/common/blocks/wooden/CircuitTableTileEntity.java
```diff
@@ -0,0 +1,328 @@
...
+	@Override
+	public EnumProperty<Direction> getFacingProperty()
+	{
+		return IEProperties.FACING_HORIZONTAL;
+	}
+
+	@Override
+	public boolean isDummy()
+	{
+		return getState().get(IEProperties.MULTIBLOCKSLAVE);
```
malte0811: This seems to be (effectively) equivalent to the default implementation?

src/main/java/blusunrize/immersiveengineering/common/blocks/wooden/LogicUnitTileEntity.java
```diff
@@ -0,0 +1,245 @@
...
+	@Override
+	public void doGraphicalUpdates(int slot)
+	{
+		this.markDirty();
+		redstoneCap.ifPresent(RedstoneBundleConnection::markDirty);
+	}
+
+	private final static int SIZE_COLORS = DyeColor.values().length;
+	private final static int SIZE_REGISTERS = LogicCircuitRegister.values().length-SIZE_COLORS;
+	EnumMap<Direction, boolean[]> inputs = new EnumMap<>(Direction.class);
```
malte0811: Variable type should probably be `Map` rather than `EnumMap`, and the field should be private (unless I've missed a usage somewhere).

src/main/java/blusunrize/immersiveengineering/common/blocks/wooden/LogicUnitTileEntity.java
```diff
@@ -0,0 +1,245 @@
...
+		return !Arrays.equals(outPre, outputs);
+	}
+
+	private void markConnectorsDirty()
+	{
+		for(Direction d : Direction.values())
+		{
+			TileEntity te = Utils.getExistingTileEntity(world, getPos().offset(d));
+			if(te instanceof ConnectorBundledTileEntity&&((ConnectorBundledTileEntity)te).getFacing()==d.getOpposite())
+				((ConnectorBundledTileEntity)te).markDirtyExtraSource();
```
malte0811: I feel like there should be some API-based way of doing this. I don't have a particular usecase for it, but IMO this is the kind of feature that should be implementable from a "clean" addon, without references to internal code.

src/main/java/blusunrize/immersiveengineering/common/blocks/wooden/LogicUnitTileEntity.java
```diff
@@ -0,0 +1,245 @@
...
+		return super.getCapability(capability, facing);
+	}
+
+	ResettableLazy<boolean[]> combinedInputs = new ResettableLazy<>(
+			() -> this.inputs.values().stream().reduce((b1, b2) -> {
+				boolean[] ret = new boolean[SIZE_COLORS];
+				for(int i = 0; i < SIZE_COLORS; i++)
+					ret[i] = b1[i]||b2[i];
+				return ret;
+			}).orElse(new boolean[SIZE_COLORS]));
```
malte0811: I think a non-`stream` solution would be more readable, something like
```java
boolean[] ret = new boolean[SIZE_COLORS];
for (boolean[] side : this.inputs.values())
    for (int i = 0; i < SIZE_COLORS; ++i)
        ret[i] |= side[i];
return ret;
```

src/main/java/blusunrize/immersiveengineering/common/blocks/wooden/LogicUnitTileEntity.java
```diff
@@ -0,0 +1,245 @@
...
+		else
+			this.registers[register.ordinal()-SIZE_COLORS] = state;
+	}
+
+	private final Map<String, VisibilityList> visibilityListMap = new ConcurrentHashMap<>();
+	private static VisibilityList visibilityTransparent = VisibilityList.show("tubes");
+
+	private String getVisibilityKey()
+	{
+		return this.inventory.stream().map(itemStack -> !itemStack.isEmpty()?"1": "0").reduce((s, s2) -> s+s2).orElse("");
```
malte0811: [`Collectors.joining()`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#joining--) would simplify this.

src/main/java/blusunrize/immersiveengineering/client/gui/CircuitTableScreen.java
```diff
@@ -0,0 +1,247 @@
...
+	private GuiButtonState<LogicCircuitRegister> outputButton;
+
+	private final ResettableLazy<Optional<LogicCircuitInstruction>> instruction = new ResettableLazy<>(() -> {
+		LogicCircuitOperator operator = getSelectedOperator();
+		if(operator==null)
+			return Optional.empty();
+		// collect inputs
+		LogicCircuitRegister[] inputs = inputButtons.stream().map(GuiButtonState::getState).filter(Objects::nonNull)
+				.limit(operator.getArgumentCount()).toArray(LogicCircuitRegister[]::new);
+		// if input array is to short, can't make an instruction
```
malte0811: to -> too. And I'd vote for splitting the stream chain into more lines.

src/main/java/blusunrize/immersiveengineering/client/gui/CircuitTableScreen.java
```diff
@@ -0,0 +1,247 @@
...
+			}
+		}
+	}
+
+	@Override
+	public void render(MatrixStack transform, int mx, int my, float partial)
+	{
+		super.render(transform, mx, my, partial);
+
+		ArrayList<ITextComponent> tooltip = new ArrayList<>();
```
malte0811: Type should be `List`, not `ArrayList`.

malte0811 (approved):
Looks good, just the minor issues you'll get in any 3.5k line PR

src/main/java/blusunrize/immersiveengineering/common/blocks/wooden/LogicUnitTileEntity.java
BluSunrize: Never knew about that one!

src/main/java/blusunrize/immersiveengineering/common/blocks/wooden/LogicUnitTileEntity.java
BluSunrize: I'm skipping on this for now, because the concept of "dirty extra source" only exists in the ConnectorBundled implementation, which is not in the API.

Unless you are suggesting a "markDirtyExternal" method in IRedstoneConnector?

src/main/java/blusunrize/immersiveengineering/client/gui/CircuitTableScreen.java
```diff
@@ -0,0 +1,256 @@
...
+
+		int stored = (int)(46*(tile.getEnergyStored(null)/(float)tile.getMaxEnergyStored(null)));
+		fillGradient(transform, guiLeft+217, guiTop+16+(46-stored), guiLeft+224, guiTop+62, 0xffb51500, 0xff600b00);
+	}
+
+	@Override
+	public boolean keyPressed(int keyCode, int scanCode, int modifiers)
+	{
+		return super.keyPressed(keyCode, scanCode, modifiers);
+	}
```
malte0811: I think this can be removed now? Unless you're overriding it to change access modifiers or some other non-obvious reason, in which case there should be a comment IMO.

src/main/java/blusunrize/immersiveengineering/client/gui/elements/GuiButtonLogicCircuitRegister.java
```diff
@@ -0,0 +1,98 @@
...
+			if(number>=0 && number<8)
+			{
+				this.setStateByInt(number+16); // plus 16 colors
+				this.onPress.onPress(this);
+				return true;
+			}
+		}
+		else if(Character.isAlphabetic(codePoint))
+		{
+			List<Integer> options = SPLIT_BY_INITIAL.get(codePoint);
```
malte0811, line 88: Maybe convert to lower case, just to be safe?

malte0811 (approved):

src/main/java/blusunrize/immersiveengineering/common/blocks/wooden/LogicUnitTileEntity.java
malte0811: After actually looking at the code my suggestion would be to have one `RedstoneBundleConnection` instance for each side, and to use the existing `markDirty`/`pollDirty` flow. I assume you didn't use it because the first connector would clear the `dirty`-flag before the others can react to it?

BluSunrize: Yep, that's exactly why that didn't work.

malte0811: What about my "one `RedstoneBundleConnection` instance per side" suggestion? If each side has its own instance everything should work correctly as far as I can see.

PR #4815, opened 2021-05-12
Change manual for feedthrough insulators ( ru and en versions )

KnewOne:
Specify the creationg process of feedthrough insulators by adding the "right click to form" part.


BluSunrize:
Thank you =)

PR #4830, opened 2021-05-26
Translate manual to German and change some context specific string

astrutz:


BluSunrize:
Can't say I like that so many jokes got lost in literal translation, but hey, at least I didn't have to write it myself.

Are you done with this, can I merge it, or do you need to do any further work on it?

astrutz:
I mean - we are talking about Germany, nothing funny about us. 😄 

I'm done with it, yes. It's quite hard to translate puns, so I decided to go with literal translations. If anyone wants to edit something in the future, it's fine.

PR #4976, opened 2021-09-18
Updated KR Translation

Kardane:
Older KR Translation is no longer matches current version, So I updated new KR translation.
Also added manual translations as well.


Kardane:
Um I missed new translation PR 
LMAO

PR #4997, opened 2021-09-27
IEOBJ refactor

malte0811:
No more OnlyIn, and hopefully clearer overall structure. As I said on discord, a lot of the changes come from moving code from one file to another.


src/api/java/blusunrize/immersiveengineering/api/tool/IDrillHead.java
```diff
@@ -81,5 +81,6 @@
 	 * This IIcon should be stitched in the item sheet
 	 */
 	@OnlyIn(Dist.CLIENT)
+	//TODO only return RL(?)
```
BluSunrize, line 84: Leaving this for a second PR?

src/main/java/blusunrize/immersiveengineering/client/models/obj/OBJHelper.java
```diff
@@ -139,6 +139,7 @@
 		}
 	}
 
+	// TODO replace with Mixin duck?
```
BluSunrize, line 142: quack?

src/main/java/blusunrize/immersiveengineering/common/wires/IEWireTypes.java
```diff
@@ -167,24 +167,14 @@ public double getRenderDiameter()
...
-					throw new IllegalStateException("Ordinal "+type+" is not valid");
-			}
+			return switch(type)
+					{
+						case COPPER, COPPER_INSULATED -> LV_CATEGORY;
+						case ELECTRUM, ELECTRUM_INSULATED -> MV_CATEGORY;
+						case STEEL -> HV_CATEGORY;
+						case STRUCTURE_ROPE, STRUCTURE_STEEL -> STRUCTURE_CATEGORY;
+						case REDSTONE -> REDSTONE_CATEGORY;
+					};
```
BluSunrize, line 177: Idealy stuff like this shouldn't have been part of this PR, but I don't thing it's a problem per se.

BluSunrize (approved):
LGTM

src/api/java/blusunrize/immersiveengineering/api/tool/IDrillHead.java
malte0811: Probably not even a PR, but essentially yes.

src/main/java/blusunrize/immersiveengineering/client/models/obj/OBJHelper.java
malte0811: A lot of OBJHelper (and reflection into MC/Forge in general) can be done with Mixins instead, which would give compile-time guarantees that things will actually work. "Duck" in this case is a common term for an interface added to a class via a Mixin (similar to how our existing accessors work). Then you can pass around `MeshAccess`s (which would be the `ModelMesh`s themselves) or similar, instead of `Object`s in a thin reflection wrapper.

PR #5001, opened 2021-09-30
Add effect descriptions. (Support for Just Enough Effect Descriptions)

BeepSterr:
Excuses if this PR has issues, I'm not a java developer but since these are only changes to language files, It should be fine.
This PR adds support for https://www.curseforge.com/minecraft/mc-mods/just-enough-effect-descriptions-jeed/

Please check if the description is accurate


BluSunrize:
Thanks!

TeamSpen210:
Might also be good to add the `jeed:effect_provider` "recipes" as mentioned there, so that the relevant items show up as producing the effects.

Issue #5027, opened 2021-10-10
[1.17.1 | #33] [Audio only] Metal Press doesn't make sound anymore


MalkContent:
As title.
Probably related to https://github.com/BluSunrize/ImmersiveEngineering/commit/cd45b7482ddc2d3e9deb954496a907affa849be8
I'll try to make a fix for this myself, but ofc I won't hog the issue

PR #5029, opened 2021-10-10
fixes metal press not playing sounds, potential for looping sound

MalkContent:
fixes metal press not playing sounds
re-added redstone check to prevent possibility of sound looping 20 times a second by toggling on redstone on a tick a sound is played
fixes https://github.com/BluSunrize/ImmersiveEngineering/issues/5027

(client side `playSound` needs to pass the client's `Player` entity instead of `null`. Interestingly sound only played before, because the code also ran on the server and broadcasted the sound, making the client play it. I learned something, I suppose, and maybe someone reading this, does, too \^\^)

side notes:

Sound is not broadcasted by the server side `playSound` anymore, which I suppose is the whole reason of putting it into the client side tick in the first place. Nothing I learned in the last 2 hours says, this should be bad in some way, but I'm eyeing it a bit critically.

Also, I was not sure if I should have added `@OnlyIn(Dist.CLIENT)` here, because the function uses LocalPlayer now but it didn't burn without it, so..

My point is, look it over if you have the time

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/MetalPressBlockEntity.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/MetalPressBlockEntity.java
@@ -12,6 +12,7 @@ import blusunrize.immersiveengineering.api.crafting.MetalPressRecipe;
 import blusunrize.immersiveengineering.api.tool.conveyor.ConveyorHandler.IConveyorAttachable;
 import blusunrize.immersiveengineering.api.utils.CapabilityReference;
 import blusunrize.immersiveengineering.api.utils.DirectionalBlockPos;
+import blusunrize.immersiveengineering.client.ClientUtils;
 import blusunrize.immersiveengineering.common.blocks.IEBlockInterfaces.IBlockBounds;
 import blusunrize.immersiveengineering.common.blocks.IEBlockInterfaces.IPlayerInteraction;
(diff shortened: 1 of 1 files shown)


XFactHD:
> Also, I was not sure if I should have added `@OnlyIn(Dist.CLIENT)` here, because the function uses LocalPlayer now but it didn't burn without it, so..

Just a comment from an outside observer:
If you can start a dedicated server, join the server and use a metal press, then you are good.
If it blows up, then you need to fix it. Adding `@OnlyIn(Dist.CLIENT)` is NOT the fix. It might appear to "fix" stuff in dev but it doesn't do anything at all in production.

MalkContent:
ty, should have been diligent enough in the first place
this is tested with running server and client in dev environment now
sound plays when being joined like that
as long as there's no jvm hijinks going on, this should be fine now

malte0811:
>  Adding `@OnlyIn(Dist.CLIENT)` is NOT the fix. It might appear to "fix" stuff in dev but it doesn't do anything at all in production.

That is not correct. `@OnlyIn` definitely works in prod, as evidenced by many large mods (IE on 1.16.5, Create, Mekanism, Botania; just to name a few) using it in the necessary places. *However* the first part of your statement is still correct in a sense: Forge is discouraging use of `@OnlyIn` (with hints that it will be removed in some future version), and in 1.17 I am not aware of any cases where it is actually necessary (in 1.16 this is not true, whatever cpw claims). I recently removed all usages of `@OnlyIn` from IE on 1.17, so please don't add any new ones.

PR #5032, opened 2021-10-11
Globally sync Water Wheel rotation

MalkContent:
synchronizes all waterwheels that are going at the same speed by setting their `rotation` to `level.getGameTime()*perTick % 1`

changing `perTick` by changing the waterflow starts a grace period where the rotation is not synced to the game time so players still perceive smooth transitions in waterwheel speed/rotation instead of janky jumping of the wheel. after the grace period the wheel makes a single jump, hopefully when noone is looking anymore.


MalkContent:
if you've got the time, check out the new smooth transition into the synced state
just wanted to see if I could do that like I had in my head
if it's to your liking, I'll clean it up, otherwise, I'll just make it so it syncs rotation when the BE is loaded and not "live", like was discussed some 5(?) months ago
sorry this took so long

MalkContent:
also: the huge list of commits makes me think that the rebasing didn't go as planned, at least on the remote side..
I'll look into that tomorrow

TeamSpen210:
You're trying to merge into 1.17, so it's including all the commits since then. Change the target branch for the PR?

MalkContent:
ty for that
didn't notice yesterday late

Issue #5030, opened 2021-10-11
[1.17.1 | #33] Waterwheels that are placed the opposite way don't get recognized by the Kinetic Dynamo


MalkContent:
The Kinetic Dynamo only recognizes Waterwheels that are placed "against it". When the wheels are placed against a block opposite to the Kinetic Dynamo's turny-thing-here-slot, they don't get recognized for powerproduction and synchronization (sync of the 3 wheels placed on a single dynamo).

This avoids a problem where two opposing Kinetic Dynamos could grab the same water wheels, but the two different facings are indistinguishable by look and can confuse.

I'll try to come up with a fix

PR #5035, opened 2021-10-14
Water wheel linking fix

MalkContent:
makes linking of the up to 3 waterwheels "more deterministic"

IRotationAcceptors can now grab wheels independent of them facing the opposite way
Opposing IRotationAcceptors now don't race anymore to grab wheels for linking, but affiliation is depending on the wheels initial facing
"Slave"-wheels don't reset to non-slaves on tick anymore, only potentially when the blocks in the link get updated

contains an api change, be advised:
IRotationAcceptor now has a method that checks if a side actually accept rotation. Previously wheels would link even if the dynamo was not oriented the correct way

fixes https://github.com/BluSunrize/ImmersiveEngineering/issues/5030

### DIFF
--- a/src/api/java/blusunrize/immersiveengineering/api/energy/IRotationAcceptor.java
+++ b/src/api/java/blusunrize/immersiveengineering/api/energy/IRotationAcceptor.java
@@ -15,4 +15,6 @@ import javax.annotation.Nonnull;
 public interface IRotationAcceptor
 {
 	void inputRotation(double rotation, @Nonnull Direction side);
+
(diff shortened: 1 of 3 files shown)


malte0811:
I've decided to implement this myself, since I don't think the "when multiple dynamos are possible, determine it by initial orientation" part is really relevant and leaving that out makes the implementation much simpler and easier to maintain. I've also moved `IRotationAcceptor` to a capability, since that's the "canonical" way of implementing facing-dependant API on BEs.

PR #5071, opened 2021-11-12
Added italian translation

maicol07:


BluSunrize:
For future reference, you do nto need to update the credits.md

And if you do, please stick to the pattern used there, which is GitHub usernames.

PR #5090, opened 2021-11-28
Adjust to 1.17's new ore system

### LINKED ISSUE #5075: [1.17.1] Feature: Raw Ores & Raw Storage
Vanilla ore blocks now drop "raw ore", e.g.: https://minecraft.fandom.com/wiki/Raw_Iron

We need to introduce this for all our ores:
- Aluminium
- Lead
- Silver
- Nickel
- Uranium

Drop rates should orient themselves on vanilla blocks, being 1 raw ore per block, except for copper, which drops 2-3 raw ore.
I don't think any of our ores need to drop 2-3.

Edit (@malte0811): We also need textures for "raw [metal] blocks" (storage blocks of raw ore), this matches vanilla.

### LINKED ISSUE #5076: [1.17.1] Feature: Deepslate Ores
Vanilla ores can now generate in deepslate, having a unique block & texture when down there:
https://minecraft.fandom.com/wiki/Deepslate

Our ores equally need deepslate variants. This requires new textures.

I'm yet unsure on whether we want aluminium to generate that deep.

### LINKED ISSUE #5091: [1.17.1] Feature: Ore Processing
With vanilla adding raw ores, we need to adjust ore processing.

I think allowing Raw Ore to be crushed into 2 dusts should be fine. Assuming we keep the "normal" crushing of Ore Blocks, this makes Fortune a far more powerful enchantment.

*Note: IE has no fortune feature. Our mining drill may lose a lot of value here.*

Similarly, one Raw Ore should smelt into 2 ingots in the arc furnace. I'm unsure on having this produce slag, my gut feeling is "no".

Edit by @malte0811: We may also want to add crushing/smelting recipes for blocks of raw ore

BluSunrize:
Adds raw ores & storage blocks, closes #5075

Adds deepslate ore, closes #5076

Adds ore processing for it all, closes #5091


BluSunrize:
I need to rebase this still >_>

BluSunrize:
Ready for review

src/datagen/java/blusunrize/immersiveengineering/data/IEBlockTags.java
```diff
@@ -95,9 +95,17 @@ protected void addTags()
...
 					tag(tags.ore).add(IEBlocks.Metals.ORES.get(metal).get());
+					tag(tags.ore).add(IEBlocks.Metals.DEEPSLATE_ORES.get(metal).get());
 					tag(Tags.Blocks.ORES).addTag(tags.ore);
+					assert tags.rawBlock!=null;
+					tag(tags.rawBlock).add(IEBlocks.Metals.RAW_ORES.get(metal).get());
 				}
 			}
+			else //TODO Forge#789, remove if/when we get tags for raw blocks
+			{
+				assert tags.rawBlock!=null;
```
malte0811: Can you replace the `assert`s with `Precondition.checkState` or in this case `Precondition.checkNonNull`? I don't think `assert`s are enabled by default, even in dev.

src/datagen/java/blusunrize/immersiveengineering/data/IEBlockTags.java
```diff
@@ -95,9 +95,17 @@ protected void addTags()
 				{
 					assert tags.ore!=null;
 					tag(tags.ore).add(IEBlocks.Metals.ORES.get(metal).get());
+					tag(tags.ore).add(IEBlocks.Metals.DEEPSLATE_ORES.get(metal).get());
```
malte0811: You should be able to just chain these two calls, so
```java
tag(tags.ore)
    .add(IEBlocks.Metals.ORES.get(metal).get())
    .add(IEBlocks.Metals.DEEPSLATE_ORES.get(metal).get());
```
That would be more readable IMO since it's obvious that both are added to the same tag.

src/datagen/java/blusunrize/immersiveengineering/data/loot/BlockLoot.java
```diff
@@ -110,6 +119,11 @@ public void accept(BiConsumer<ResourceLocation, Builder> out)
 			ConveyorBlock block = entry.get();
 			register(entry, singleItem(block).apply(ConveyorCoverLootFunction.builder()));
 		}
+		for(EnumMetals metal : EnumMetals.values())
+		{
+			if(metal.shouldAddOre())
+				registerOre(metal);
+		}
```
malte0811: Superfluous parenthesis

malte0811 (changes requested):
I think you're still missing worldgen for deepslate ores. You'll probably want to store a list of `TargetBlockState` in `IEOreFeatureConfig` instead of `target` and `size` and then adapt the rest of the code accordingly.

src/datagen/java/blusunrize/immersiveengineering/data/IEBlockTags.java
BluSunrize: What do you mean `assert`s are not enabled? Are you telling me the common Java keyword `assert` doesn't work in Minecraft?! O_o

malte0811: Less MC and more Java in general, at least according to https://stackoverflow.com/questions/29120928/why-is-assert-not-enabled-by-default-in-java

BluSunrize: As someone who uses Python primarily these days: What the fuck.
I mean, I get their approach, but wow that is so unlike what I'd expect...

BluSunrize:
Changes made, ready for re-review.

BluSunrize:
Nevermind, forgot the ore generation.

BluSunrize:
Deepslate ores are now in, other changes have been postponed to 1.18.

Ready for review.

malte0811 (approved):

PR #5148, opened 2021-12-29
1.18.1 "Hammer" to "Screwdriver"

Vapaman:
Also a minor typo in external heaters entry

I also did the 1.16.5 version

PR #5159, opened 2022-01-02
Fix manual references(as in #5148)

Smollet777:


BluSunrize:
Okay this one can't be switched to 1.18.2

If this is still a relevant change, please rebase it to 1.18.2, since that is our development version.

PR #5165, opened 2022-01-04
New copper colors for all the copper-colored textures I could find

voidsong-dragonfly:
I added a few more since I last suggested them, including the dynamo (copper coil) and the LV coil itself.


BluSunrize:
Thanks ^_^

PR #5169, opened 2022-01-06
Move crates to vanilla BE superclass

malte0811:
Also makes the general "infrastructure" related to crates less dependant on IEs BE classes.


src/main/java/blusunrize/immersiveengineering/client/ClientEventHandler.java
```diff
@@ -534,14 +533,10 @@ else if(equipped.getItem() instanceof IEShieldItem)
 					Entity entity = ((EntityHitResult)mop).getEntity();
 					if(entity instanceof ItemFrame)
 						BlockOverlayUtils.renderOreveinMapOverlays(transform, (ItemFrame)entity, mop, scaledWidth, scaledHeight);
-					else if(entity instanceof IEMinecartEntity)
+					else if(entity instanceof IEMinecartEntity<?> ieMinecart&&ieMinecart.getContainedBlockEntity() instanceof IBlockOverlayText overlayText)
 					{
```
BluSunrize, line 537: Just have to say again, I *love* the new casting feature.

BluSunrize (approved):
LGTM.

Is the way to define loot tables still the same? I didn't look into the NBT file changes.

malte0811:
> Is the way to define loot tables still the same?

The key changed from `lootTable` to `LootTable` (since that's what vanilla uses), other than that it's the same as before. The old key will still work, it's automatically converted when loading the BE.

Issue #5002, opened 2021-09-30
[Request][1.17.1 | #24] Turn Voltmeter into a Multimeter that can also read redstone strength


MalkContent:
because it'd be really handy to be able to read the strength off of redstone dust and redstone connectors

PR #5215, opened 2022-02-06
Multimeter

MalkContent:
added Engineer's Voltmeter functionality:
-hover-over readout of (highest) redstone level for redstone signal sources (text display near crosshair)
-shift-click of any block to read out redstone level (chat display)

renamed Engineer's Voltmeter to Engineer's Multimeter (lang file only)

does NOT contain manual text besides renaming it in the wiring entry

see https://github.com/BluSunrize/ImmersiveEngineering/issues/5002


BluSunrize:
Why does this need packets?

MalkContent:
without the packets the client can have inaccurate information
for example the readout of a comparator wouldn't display properly
I asked about it before I added that, though that's quite some time ago now

Issue #5281, opened 2022-04-30
[1.18.2] Incorrect model rotation for post transformer, current transformer and razor wire


alanbriolat:
### Description of the issue:
The models of post transformer, current transformer and razor wire are rendered 180 degrees rotated from their expected appearance:

![2022-04-30_20 35 15](https://user-images.githubusercontent.com/12193/166120501-a66f5dd4-960c-47af-951b-63f3b8f0dfd9.png)

- Post transformer is floating in mid-air (but as #5243 mentions, collision box and attachment points are where you'd expect them to be if the model was rendered in the right place, and therefore also floating in mid-air)
- Current transformer dial is facing away from the player when placed, not towards
- Razor wire wooden ends do not appear on the ends, but in the middle instead

With some digging, 66708713c35c0f9e0da33c9e599852c3c1d8fddb ("Simplify connector blockstate generation code") stood out as a likely culprit, and indeed when I test at that commit I get the same behaviour as above, but if I test the commit before it, I get the expected behaviour, included here for comparison:

![2022-04-30_20 32 34](https://user-images.githubusercontent.com/12193/166120518-abd6f6c1-e86e-47b2-ab36-de7fd55f76fd.png)

(This is true even if re-running `./gradlew runData`, so it's not just the change in the generated data in those 2 commits, it's actually being generated with different values for rotation.)

I was hoping to submit a pull request instead of a bug report, but I'm not really familiar with Minecraft modding, and I haven't yet figured out what is different about those blocks vs. the other 8 using `createHorizontalRotatedBlock` - I tested all the others and they weren't noticeably changed by that commit. I'm happy to try and implement a fix if pointed in the right direction.

(This is technically a duplicate of #5243, however I'd suggest #5243 is now a duplicate of this one, since it covers only part of the problems introduced by that refactor commit.)

### Crashlog:
(Not a crashing bug.)

### Versions & Modlist
- Minecraft 1.18.2
- Forge 40.1.0
- Immersive Engineering 1.18.2-8.0.1-147 (also confirmed in latest `1.18.2` branch, revision a4ead70fe1e6d26c9698fda482f7c1bb48373fe2)

Bug is reproducible with just the above, however rest of mod list included for completeness:

- Configured 1.5.3
- Cloth Config v4 API 6.2.57
- Xaero's Minimap 22.6.0
- Xaero's World Map 1.20.6
- Refined Storage 1.10.2
- Architectury 4.2.50
[...]

PR #5282, opened 2022-04-30
Fix rotation of post transformer, current transformer and razor wire

alanbriolat:
A proposed fix for #5281, but I'm open to suggestions. I don't know my way around this code or Minecraft modding in general, so I'm not sure how you would want to fix this problem, I just wanted to demonstrate at least one viable solution to satisfy my own curiosity.

From what I could tell, the difference introduced by 66708713c35c0f9e0da33c9e599852c3c1d8fddb for these 3 models was that previously they didn't go through `createRotatedBlock(block, model, additionalProps)`, and that method applies a 180 degree rotation. By unifying things via `createHorizontalRotatedBlock` those 3 models gained the 180 degree rotation (but there was no change for the other 8 models that use `createHorizontalRotatedBlock`, because they used `createRotatedBlock` before).


malte0811:
Looks good, thanks!

PR #5311, opened 2022-05-31
Export item icons and tag contents during datagen

malte0811:
The goal is to use these in the online version of the manual. Only icons for IE items are exported, for vanilla items we have to check if it would be a copyright violation.

The icons and tags are exported to the directory specified in `$ie_manual_datagen_path`, or not at all if it is not specified. In GH actions on version branches (matching `1.*`), they are pushed to the branch `manual-data/$(git branch --show-current)`. Since this code pretty much only runs in GH actions, it does not have to consider other mods and can be relatively "brittle"/hacky.


BluSunrize (approved):
LGTM

PR #5337, opened 2022-06-21
Add IMultiblock:getBlock

sguest:
Allow addons to get a multiblock's block-item.

Adding to `IMultiblock` as the API looks like it encourages use of that interface. Other idea would be add it to `TemplateMultiblock` since currently those are the only ones that will have a non-air block.


BluSunrize:
Sure, why not

PR #5354, opened 2022-07-09
Fixed a small typo in both issue templates

TwistedGate:
I am assuming these two are typos, simply because "size" does not make sense :D


BluSunrize:
Hah, whoops!

Well spotted!

PR #5358, opened 2022-07-16
update oc2 version

leedagee:
oc2 has changed its device registry name from `block_device_providers` to `block_device_provider` and changed some api method signature, thus a update of oc2 version is needed.


geferon:
Wouldn't it be better to use the Registries class in Utils instead of manually providing the ResourceLocation?
Like this https://github.com/geferon/ImmersiveEngineering/commit/b8b03e10ae7997e49a5fce16405c6d0cd70be044

PR #5360, opened 2022-07-17
Add IMPLY and NIMPLY logic operators

sashafiesta:
IMPLY operator: https://en.wikipedia.org/wiki/IMPLY_gate
NIMPLY operator: https://en.wikipedia.org/wiki/NIMPLY_gate


BluSunrize:
Good stuff, thanks for that!

PR #5379, opened 2022-07-26
Overhaul the Advancements

BluSunrize:
Most of this is generated files.
The biggest overhaul is in the advancement generation, we now have a nicer builder for those and use clean criteria triggers where possible.

I've thrown out some of the boring crafting advancements (like "Craft a Skyhook") and replaced them with more intersting stuff:
- 1km travel on the skyhook is back
- Growing Chorus in the Cloche
- A new secret advancement that involves wires

Also advancements are split into three different pages now: Main, Multiblocks, Tools.


Tomatoh:
Please add multiblock advancements for the rest of the multiblocks

BluSunrize:
I only added the ones I felt were good for progression. I don't want to swamp people in a load of advancements.

If you want them, you can add them yourself with a datapack though!

PR #5396, opened 2022-08-04
update ja_jp.json

karakufire:


karakufire:
commits included in #5401

PR #5404, opened 2022-08-06
Fix duplication of buckets in the Assembler

### LINKED ISSUE #5382: [1.18+] Redstone acid recipe in Assembler produces buckets from just redstone and water
### Minecraft Version

1.19

1.19-8.1.0-pre.68

### Expected Behavior

Assembler should not produce buckets of redstone acid from just redstone and water, or should not produce empty buckets if running recursive mode with accumulator recipe

### Actual Behavior

Assembler makes buckets of redstone acid with no iron input, therefore it makes buckets seemingly out of thin air

### Steps to reproduce

1. Create an assembler multiblock
2. Create an infinite water source with a pump and a small water pond
3. Create a power source for the assembler and the pump
4. Pump the water to the assembler
5. Add the redstone acid recipe to the assembler
6. Add redstone
7. Watch it produce (buckets) of redstone acid

### Debug Crash Log

https://gist.github.com/RRyyas/ce0851b4b73b2f8871a0641ee7132fa0

BluSunrize:
By making it aware of the lack of remaining items.

Closes #5382

@malte0811 I definitely need your feedback on this.

Technically this shouldn't result in a major version increase because it isn't API breaking? `specialQueryConverters` was removed but it was inaccessible anyway and had no method to add to it.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/IEContent.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/IEContent.java
@@ -208,7 +208,7 @@ public class IEContent
 			if(!(potentialBucket instanceof BucketItem))
 				return null;
 			// bucket was consumed in recipe
-			if(remain.isEmpty())
+			if(remain.getItem()!=Items.BUCKET)
 				return null;
 			//Explicitly check for vanilla-style non-dynamic container items
 			//noinspection deprecation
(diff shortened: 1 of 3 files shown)


src/main/java/blusunrize/immersiveengineering/common/IEContent.java
```diff
@@ -199,14 +199,17 @@ public static void init(ParallelDispatchEvent ev)
...
+		AssemblerHandler.registerSpecialIngredientConverter((o, remain) ->
 		{
 			final ItemStack[] matching = o.getItems();
 			if(!o.isVanilla()||matching.length!=1)
 				return null;
 			final Item potentialBucket = matching[0].getItem();
 			if(!(potentialBucket instanceof BucketItem))
 				return null;
+			// bucket was consumed in recipe
+			if(remain.isEmpty())
```
malte0811: Might be better to explicitly check for an empty bucket here? Probably won't happen in practice, but just in case some mod returns something that isn't a bucket. Same in the milk check below.

malte0811 (approved):
LGTM

src/main/java/blusunrize/immersiveengineering/common/IEContent.java
BluSunrize: Fair

PR #5417, opened 2022-08-26
Nether Vein Changes & Additions

voidsong-dragonfly:
- New nether veins of various types!
     - One for new soulsand based vein, including soul soil, soulsand, and gravel
     - One for magma-related blocks, including magma blocks, smooth basalt, and obsidian
 - The Archaic Digsite vein now has a more-fitting blockset
     - Removed gravel, instead added polished blackstone & bricks
     - Added gilded blackstone to the vein at the lowest priority


BluSunrize:
This needs an update with spoils now merged

BluSunrize:
Oh also, the Silt vein might need an overhaul, since gravel is a spoil now?
Alternatively, should we maybe not have spoil on silt?

voidsong-dragonfly:
My thought is something like having dirt replace it, maybe? Sand + Clay + Dirt instead of gravel? Makes the vein less directly-useful for concrete, though.

src/datagen/java/blusunrize/immersiveengineering/data/Recipes.java
```diff
@@ -1232,12 +1232,32 @@ private void mineralMixes(@Nonnull Consumer<FinishedRecipe> out)
 				.build(out, toRL("mineral/mephitic_quarzite"));
 		MineralMixBuilder.builder(nether)
 				.addNetherSpoils()
-				.addOre(Blocks.GRAVEL, .6f)
-				.addOre(Blocks.ANCIENT_DEBRIS, .4f)
+				.addOre(Blocks.POLISHED_BLACKSTONE_BRICKS, .4f)
+				.addOre(Blocks.POLISHED_BLACKSTONE, .3f)
+				.addOre(Blocks.ANCIENT_DEBRIS, .2f)
+				.addOre(Blocks.GILDED_BLACKSTONE, .1f)
```
BluSunrize, line 1238: I don't love the reduction in Ancient Debris output here.
Is there a reason other than "making space for everything else"? I must admit, I have not *played* with it, so maybe it produces ridiculous amounts of Ancient Debris?

BluSunrize (approved):
LGTM, but one question on balancing

src/datagen/java/blusunrize/immersiveengineering/data/Recipes.java
voidsong-dragonfly: Because, when I was playing around, the ancient debris haul was still _huge_ - that 0.5 instead of 0.7 on fail chance actually means it's closer to 0.7x instead of 0.5x the amount of debris. It was still a huge amount for me, especially compared to the rarity of debris in the nether.

The below image is two minutes (approximate) of runtime on an archaic digsite of 99.99% saturation
image

PR #5448, opened 2022-09-29
Fix problems with CC: Tweaked compatibility and diesel generators and…

### LINKED ISSUE #5402: Not able to use method setEnabled() with Diesel Generator in CC: Tweaked (ComputerCraft)
### Minecraft Version

1.18

8.1.0-150

### Expected Behavior

The expected behavior would be to be able enable or disable the Diesel Generator through a lua script in ComputerCraft after wrapping it as a peripheral:

```lua
lua> g = peripheral.wrap("diesel_generator_0")
lua> g.setEnabled(true) -- Enables generator.
lua> g.setEnabled(false) -- Disables generator.
lua>
```

### Actual Behavior

When using the method, an exception is thrown instead: 

```lua
lua> g = peripheral.wrap("diesel_generator_0")
lua> g.setEnabled(true) -- Enables generator.
lua> g.setEnabled(false) -- Disables generator.
lua:1: class blusunrize.immersiveengineering.common.blocks.metal.DieselGeneratorBlockEntity cannot be cast to class blusunrize.immersiveengineering.common.blocks.generic.PoweredMultiblockBlockEntity (blusunrize.immersiveengineering.common.blocks.metal.DieselGeneratorBlockEntity and blusunrize.immersiveengineering.common.blocks.generic.PowerdMultiblockBlockEntity are in module immersiveengineering@1.18.2-8.1.0-150 of loader 'TRANSFORMER' @15c6027d)
lua> 
```
Bildschirmfoto 2022-08-06 um 09 43 09

### Steps to reproduce

1. Connect a Computer from CC: Tweaked to the Redstone interface of a diesel generator using two wired modems networking cable
2. Certify that both devices are in the network. Both modems shows red "ring".
3. Right click computer to open the interface.
4. Open the Lua interactive shell, by typing `lua` and pressing [ENTER].
5. Wrap the diesel as a peripheral using the `peripheral.wrap()` method of CC: Tweaked as seen above and use the `setEnabled()`method.
6. Exception is thrown.

### Debug Crash Log

https://pastebin.com/cRDvjbgD

pespinho:
Fix problems with CC: Tweaked compatibility and diesel generators and add the new computer callable method getEnabled()

Fixed a problem causing Lua exceptions when methods isRunning() and setEnabled() were called for a diesel generator by:
- Correcting the parameter type for setEnabled() in MultiblockCallbacks.java.
- Moving isRunning() to PoweredMBCallbacks.java.
- Adding a specific version of isRunning() for diesel generators in DieselGenCallbacks.java.

Added a method getEnabled() to MultiblockCallbacks.java so that the enabled status can be also queried.

Made some changes for setting the RS block as CallbackEnvironment for Computer Calls of MultiblockCallbackOwners:
- changed preprocess() accordingly, so that it returns the RS block entity.
- changed the isRunning() method, so that it first retrieves the master block before querying.

Fixes #5402

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/util/compat/computers/generic/owners/MultiblockCallbackOwner.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/util/compat/computers/generic/owners/MultiblockCallbackOwner.java
@@ -27,10 +27,6 @@ public abstract class MultiblockCallbackOwner<T extends MultiblockPartBlockEntit
 	@Override
 	public T preprocess(T arg)
 	{
-		T master = (T)arg.master();
-		if(master!=null)
-			return master;
-		else
-			return arg;
+		return arg;
 	}
 }
--- a/src/main/java/blusunrize/immersiveengineering/common/util/compat/computers/generic/impl/MultiblockCallbacks.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/util/compat/computers/generic/impl/MultiblockCallbacks.java
@@ -20,7 +20,7 @@ public class MultiblockCallbacks extends Callback<MultiblockPartBlockEntity<?>>
 	@ComputerCallable
 	public boolean getEnabled(CallbackEnvironment<MultiblockPartBlockEntity<?>> env)
 	{
-		return env.object().computerControl.isEnabled();
+		return env.beforePreprocess().computerControl.isEnabled();
 	}
 
 	@ComputerCallable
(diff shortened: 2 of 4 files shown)


src/main/java/blusunrize/immersiveengineering/common/util/compat/computers/generic/owners/MultiblockCallbackOwner.java
```diff
@@ -27,10 +27,6 @@ public boolean canAttachTo(T candidate)
 	@Override
 	public T preprocess(T arg)
 	{
-		T master = (T)arg.master();
-		if(master!=null)
-			return master;
-		else
-			return arg;
+		return arg;
```
malte0811: This change breaks nearly every multiblock computer method, they expect to run on the master BE. I've fixed this part of the issue in the main 1.18.2 branch by passing both the master and the "direct" BE, so that computer methods can choose which one to use.

malte0811 (changes requested):

src/main/java/blusunrize/immersiveengineering/common/util/compat/computers/generic/owners/MultiblockCallbackOwner.java
pespinho: Thanks, Malte! The way I did it was indeed a workaround, by getting the master block in the computer method itself. I will rebase and review my changes.

Issue #5452, opened 2022-10-02
Revolver modifiers from the Hammer and Barrel aren't applied.


Darkally33:
### Minecraft Version

1.19

### Immersive Engineering Version

1.19.2-9.0.0-153

### Expected Behavior

Crafting a Revolver with with modifiers on the Revolver Drum, Revolver Hammer, and Revolver Barrel should result in a Revolver with the modifiers of the three parts combined.

### Actual Behavior

Only the modifiers on the Revolver Drum are applied to the resulting Revolver.
Swapping out the Hammer or Barrel for parts with different modifiers has no effect on the crafting result.

To test if this was just a tooltip issue, I tested the below Revolver against one with no modifiers, and there appeared to be no difference in fire rate.
![2022-10-02_14 29 19](https://user-images.githubusercontent.com/78669103/193477835-285e0bb2-9f66-40be-8c2a-0e08e2a5503d.png)
![2022-10-02_14 29 22](https://user-images.githubusercontent.com/78669103/193477841-e7857d18-88e4-4c2f-b460-db328adf31c3.png)
![2022-10-02_14 29 24](https://user-images.githubusercontent.com/78669103/193477843-cbf913de-6761-4ed1-b48a-47d4ea5066cd.png)
![2022-10-02_14 29 28](https://user-images.githubusercontent.com/78669103/193477846-aa4da94e-cfba-4fb9-8607-787dde2129db.png)

### Steps to reproduce

1. Spawn in and level up a bunch of Gunsmith villagers to get a selection of Revolver parts with modifiers.
2. Craft a Revolver using said parts, paying attention to what modifier each part has.
3. The resulting Revolver will only have the modifiers from the Drum.

### Debug Crash Log

https://pastebin.com/gMK7FQEp

gorberto:
This bug also exists in version 1.18.2, I've opened PRs against both versions to fix it.

BluSunrize:
Closed by merged PR, will be resolved by next release

PR #5465, opened 2022-10-16
Fix Revolver Perk Bug

gorberto:
What this PR does:
- Updates the indexes of items to check in the crafting matrix when applying perks to the revolver

Address https://github.com/BluSunrize/ImmersiveEngineering/issues/5452

### Items with Perks
barrel
drum
hammer
### Before (only the drum perks are applied)
before
### After (all perks are applied)
after


malte0811:
The file you changed is automatically generated and will be overwritten the next time datagen runs (e.g. in the GH action runs once they complete). Instead you need to change the code that generates the file, in this case [here](https://github.com/BluSunrize/ImmersiveEngineering/blob/479f02cd521a38b5e83921653f3a48e04d355915/src/datagen/java/blusunrize/immersiveengineering/data/Recipes.java#L2931).

gorberto:
> The file you changed is automatically generated and will be overwritten the next time datagen runs (e.g. in the GH action runs once they complete). Instead you need to change the code that generates the file, in this case [here](https://github.com/BluSunrize/ImmersiveEngineering/blob/479f02cd521a38b5e83921653f3a48e04d355915/src/datagen/java/blusunrize/immersiveengineering/data/Recipes.java#L2931).

Thanks for the explanation @malte0811!

BluSunrize:
You also then need to execute runData and commit the changes to the generated files.
Otherwise our CI breaks.

gorberto:
> You also then need to execute runData and commit the changes to the generated files. Otherwise our CI breaks.

Sorry about that and thank you for the explanation 😅 I'll fix the PR later this evening!

BluSunrize:
The `src/generated/resources/.cache/cache` file has a massive diff.
Is there a change between line endings maybe?

gorberto:
Yeah I noticed that too in [9f19e34](https://github.com/BluSunrize/ImmersiveEngineering/pull/5465/commits/9f19e3403244d7c1425d1a15004f9b5b05f93a81), but I thought I fixed that in [4836f17](https://github.com/BluSunrize/ImmersiveEngineering/pull/5465/commits/4836f172ec7fda3fa6b75c52243485c20e47d5c1) 🙁 Do you still see the large diff somewhere? `src/generated/resources/.cache/cache` only shows one update in the latest commit and the GH actions passed.

PR #5466, opened 2022-10-16
Fix Revolver Perk Bug (1.19.2)

gorberto:
What this PR does:
- Updates the indexes of items to check in the crafting matrix when applying perks to the revolver

Address https://github.com/BluSunrize/ImmersiveEngineering/issues/5452

### Items with Perks
barrel
drum
hammer
### Before (only the drum perks are applied)
before
### After (all perks are applied)
after


malte0811:
Our current workflow for maintaining both 1.18 and 1.19 is that changes relevant to both versions are made on 1.18, and then whole 1.18 branch is merged into 1.19  regularly. I'll close this PR and put further comments on #5465.

PR #5471, opened 2022-10-21
Fix Diesel Generator Infinite Power

### LINKED ISSUE #5470: [1.18.2] Diesel Generator runs infinite power supply with just initial fuel bucket
### Minecraft Version

1.18

8.2.2-154

### Expected Behavior

Diesel Generator stop working once it completely runs out of fuel.

### Actual Behavior

Diesel generator runs infinitely without fuel supply once initiated and enabled.
https://i.gyazo.com/67bf5d77b77bf60a54eaead4e1e31a95.mp4

(Power gauge with measuring tool has it run down to 4kFE then "reset" back to 8kFE looped)

### Steps to reproduce

1. Create diesel generator multiblock, add a pipe and add a barrel
2. Connect generator to powergrid with constant pull (excavator in this case)
2. Fill barrel with a bucket of fuel
3. Break the pipe connecting the barrel of fuel and the generator and watch it keep supplying power indefinitely.

Screenshot_44
Screenshot_45

Reproduced with a clean install of IE in creative
Screenshot_46

### Debug Crash Log

https://pastebin.com/F2gDGZ8Z

gorberto:
What this PR does:
- Prevents the diesel generator from running when the amount of fuel in the tank is less than the minimum amount of fuel required to burn 

Closes https://github.com/BluSunrize/ImmersiveEngineering/issues/5470

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/DieselGeneratorBlockEntity.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/DieselGeneratorBlockEntity.java
@@ -159,7 +159,7 @@ public class DieselGeneratorBlockEntity extends MultiblockPartBlockEntity<Diesel
 		tickCommon();
 		final boolean prevActive = active;
 
-		if(!isRSDisabled()&&!tanks[0].getFluid().isEmpty())
+		if(shouldRunGenerator())
 		{
 			int output = IEServerConfig.MACHINES.dieselGen_output.get();
 			List<IEnergyStorage> presentOutputs = outputs.stream()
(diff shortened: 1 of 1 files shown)


src/main/java/blusunrize/immersiveengineering/common/blocks/metal/DieselGeneratorBlockEntity.java
```diff
@@ -172,27 +172,13 @@ public void tickServer()
 			}
 			if(consumeTick<=0) //Consume 10*tick-amount every 10ticks to allow for 1/10th mB amounts
 			{
-				GeneratorFuel recipe = recipeGetter.apply(level, tanks[0].getFluid().getFluid());
-				if(recipe!=null)
+				if (!active)
 				{
-					int burnTime = recipe.getBurnTime();
-					int fluidConsumed = (10*FluidAttributes.BUCKET_VOLUME)/burnTime;
-					if(tanks[0].getFluidAmount() >= fluidConsumed)
```
gorberto, line 180: When the amount of fluid in the tank is less than `fuelConsumed`, but is not empty (thus passing the check on line 162), the remaining fuel is not drained. This allows the generator to output flux (line 169) indefinitely.

BluSunrize:
I opted for a different, smaller fix:

https://github.com/BluSunrize/ImmersiveEngineering/commit/b6489825f0cc6d3c1fd2936b856649335f9edf1d

PR #5479, opened 2022-11-14
Crusher, Sample Drill, Excavator Particle Tweaks

voidsong-dragonfly:
- Increase particle output of excavator to make it look like a full block is being dumped
 - Give sample drill cobblestone particles for digging
 - Add in crusher crushing particles of the kind of thing it's currently chewing through

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/CrusherBlockEntity.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/metal/CrusherBlockEntity.java
@@ -92,18 +92,22 @@ public class CrusherBlockEntity extends PoweredMultiblockBlockEntity<CrusherBloc
 	public void tickServer() {
 		super.tickServer();
 		if (processQueue.size() > 0 && processQueue.get(0).getRecipe(getLevelNonnull()) != null)
-		{
 			spawnParticles(processQueue.get(0).getRecipe(getLevelNonnull()).input.getItems()[0]);
-		}
 	}
 
 	public void spawnParticles(ItemStack stack)
 	{
 		if (getLevel() instanceof ServerLevel serverLevel)
 		{
-			serverLevel.sendParticles(new ItemParticleOption(ParticleTypes.ITEM, stack), getBlockPos().getX() + (getFacing().getAxis().equals(Axis.X) ? 0.5 : 0), getBlockPos().getY() + 2.125, getBlockPos().getZ() + 0.5 -(getFacing().getAxis().equals(Axis.X) ? 0.5 : 0) , 8, 0, 0, 0, 0.0625);
-			serverLevel.sendParticles(new ItemParticleOption(ParticleTypes.ITEM, stack), getBlockPos().getX() + 0.5, getBlockPos().getY() + 2.125, getBlockPos().getZ() + 0.5, 8, 0, 0, 0, 0.0625);
-			serverLevel.sendParticles(new ItemParticleOption(ParticleTypes.ITEM, stack), getBlockPos().getX() + 1 - (getFacing().getAxis().equals(Axis.X) ? 0.5 : 0), getBlockPos().getY() + 2.125, getBlockPos().getZ() + 0.5 + (getFacing().getAxis().equals(Axis.X) ? 0.5 : 0), 8, 0, 0, 0, 0.0625);
+			serverLevel.sendParticles(new ItemParticleOption(ParticleTypes.ITEM, stack),
+				getBlockPos().getX() + (getFacing().getAxis().equals(Axis.X) ? 0.5 : 0), getBlockPos().getY() + 2.125, getBlockPos().getZ() + 0.5 -(getFacing().getAxis().equals(Axis.X) ? 0.5 : 0),
+				8, 0, 0, 0, 0.0625);
+			serverLevel.sendParticles(new ItemParticleOption(ParticleTypes.ITEM, stack),
+				getBlockPos().getX() + 0.5, getBlockPos().getY() + 2.125, getBlockPos().getZ() + 0.5,
+				8, 0, 0, 0, 0.0625);
+			serverLevel.sendParticles(new ItemParticleOption(ParticleTypes.ITEM, stack),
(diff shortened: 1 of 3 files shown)


src/main/java/blusunrize/immersiveengineering/common/blocks/metal/CrusherBlockEntity.java
```diff
@@ -84,6 +88,17 @@ public void tickClient()
 		}
 	}
 
+	public void spawnParticles(ItemStack stack)
+	{
+		Level level = getLevelNonnull();
+		if(level instanceof ServerLevel)
```
malte0811: I'd recommend using pattern-matching instanceof here (`if (getLevel() instanceof ServerLevel serverLevel)`)

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/CrusherBlockEntity.java
```diff
@@ -84,6 +88,17 @@ public void tickClient()
...
+		{
+			((ServerLevel)level).sendParticles(new ItemParticleOption(ParticleTypes.ITEM, stack), getBlockPos().getX(), getBlockPos().getY() + 2.125, getBlockPos().getZ() + 0.5 , 8, 0, 0, 0, 0.0625);
+			((ServerLevel)level).sendParticles(new ItemParticleOption(ParticleTypes.ITEM, stack), getBlockPos().getX() + 0.5, getBlockPos().getY() + 2.125, getBlockPos().getZ() + 0.5, 8, 0, 0, 0, 0.0625);
+			((ServerLevel)level).sendParticles(new ItemParticleOption(ParticleTypes.ITEM, stack), getBlockPos().getX() + 1, getBlockPos().getY() + 2.125, getBlockPos().getZ() + 0.5, 8, 0, 0, 0, 0.0625);
```
malte0811: Did you test this with different crusher orientations? It looks like this will only work for north/south facing crushers. Also I'd split up those lines a bit.

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/CrusherBlockEntity.java
```diff
@@ -342,6 +357,10 @@ public boolean isInWorldProcessingMachine()
 	@Override
 	public boolean additionalCanProcessCheck(MultiblockProcess<CrusherRecipe> process)
 	{
+		if (processQueue.size() > 0 && processQueue.get(0).getRecipe(getLevelNonnull()) != null)
+		{
+			spawnParticles(processQueue.get(0).getRecipe(getLevelNonnull()).input.getItems()[0]);
+		}
```
malte0811: This doesn't seem like the right place for spawning particles. I'd suggest overriding `tickServer` and spawning the particles from there.

malte0811 (changes requested):

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/CrusherBlockEntity.java
voidsong-dragonfly: .... I _may_ have forgotten that I made that temporary. Thanks for catching that!

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/CrusherBlockEntity.java
voidsong-dragonfly: Should be fixed in newest commit

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/CrusherBlockEntity.java
```diff
@@ -84,6 +88,25 @@ public void tickClient()
...
+	}
+
+	public void spawnParticles(ItemStack stack)
+	{
+		if (getLevel() instanceof ServerLevel serverLevel)
+		{
+			serverLevel.sendParticles(new ItemParticleOption(ParticleTypes.ITEM, stack), getBlockPos().getX() + (getFacing().getAxis().equals(Axis.X) ? 0.5 : 0), getBlockPos().getY() + 2.125, getBlockPos().getZ() + 0.5 -(getFacing().getAxis().equals(Axis.X) ? 0.5 : 0) , 8, 0, 0, 0, 0.0625);
+			serverLevel.sendParticles(new ItemParticleOption(ParticleTypes.ITEM, stack), getBlockPos().getX() + 0.5, getBlockPos().getY() + 2.125, getBlockPos().getZ() + 0.5, 8, 0, 0, 0, 0.0625);
+			serverLevel.sendParticles(new ItemParticleOption(ParticleTypes.ITEM, stack), getBlockPos().getX() + 1 - (getFacing().getAxis().equals(Axis.X) ? 0.5 : 0), getBlockPos().getY() + 2.125, getBlockPos().getZ() + 0.5 + (getFacing().getAxis().equals(Axis.X) ? 0.5 : 0), 8, 0, 0, 0, 0.0625);
```
BluSunrize: Can we get some linebreaks in here? You did it for the SampleDrill, I think they'd improve readabilitiy here too.

src/main/java/blusunrize/immersiveengineering/common/blocks/metal/CrusherBlockEntity.java
```diff
@@ -84,6 +88,25 @@ public void tickClient()
 		}
 	}
 
+	@Override
+	public void tickServer() {
+		super.tickServer();
+		if (processQueue.size() > 0 && processQueue.get(0).getRecipe(getLevelNonnull()) != null)
+		{
```
BluSunrize: Brackets here are superfluous.

BluSunrize (commented):
Just a few cosmetic things

TeamSpen210:
Found a little bug - the crusher is using the input item from the recipe, not the item actually tossed in. So if you toss in deepslate or nether ores, you can see some stone particles since it's using the overworld block.

Edit: A little digging suggests the issue might be with `IMultiblockRecipe.getDisplayStack()` - it just returns a sized copy of the first matching item in its ingredient list, which may not match the actual item. Is that intentional? That's what's saved in the `MultiblockProcessInWorld`, meaning the actual original input is discarded entirely.

voidsong-dragonfly:
To the crusher, that _is_ the item tossed in. I don't have a record of the actual item tossed in, just the recipe, and storing recipe makes more sense than storing item, I think.

PR #5480, opened 2022-11-14
Arc Furnace & Blast Furnace Smoke

voidsong-dragonfly:
- Adds arc furnace textures using blast brick
 - Adds arc furnace smoke particles + spark particles, dependent upon the load of the furnace
 - Adds blast furnace smoke particles to work along the arc furnace ones


BluSunrize:
Why is there a massive diff for the model itself?
What changed there to double its linecount?

voidsong-dragonfly:
I am very unsure, I believe I just changed a few UVs and then Blender did _that_.
I don't think vertex order is quite formulaic?

malte0811:
>  What changed there to double its linecount?

At a guess, did you run `obj_simplify.py` on the new OBJ file?

BluSunrize:
> I am very unsure, I believe I just changed a few UVs and then Blender did _that_. I don't think vertex order is quite formulaic?

Out of interest, what UV changes were necessary and why?
And yes, Malte is quite right, if you *can* (if you have a python setup) please run obj_simplify on it.
If not, we'll have to do it after merge.

voidsong-dragonfly:
The changes I made were a few to make the inside of the arc furnace cauldron be brick while he outside was steel. Otherwise, I should have made no changes

BluSunrize:
This looks fine now. It's a massive change because the order of fields has changed, but that's fine. I just didn't want double the linecount =P

PR #5481, opened 2022-11-14
More VFX Updates

voidsong-dragonfly:
- Sawmill particles now obey particle density setting & blade stops when RS is disabled and it has nothing in queue
 - Diesel gen particles now obey particle density setting, and are twice as dense
 - Mixer particles are significantly more common & water level stops one pixel below the top of the vat for a little more pizzaz

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/client/render/tile/SawmillRenderer.java
+++ b/src/main/java/blusunrize/immersiveengineering/client/render/tile/SawmillRenderer.java
@@ -60,7 +60,7 @@ public class SawmillRenderer extends IEBlockEntityRenderer<SawmillBlockEntity>
 			matrixStack.pushPose();
 			matrixStack.translate(1, .125, -.5);
 			float spin = te.animation_bladeRotation;
-			if(te.shouldRenderAsActive())
+			if(te.shouldRenderAsActive() && te.sawmillProcessQueue.size() > 0)
 				spin += 36f*partialTicks;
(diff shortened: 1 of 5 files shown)


src/main/java/blusunrize/immersiveengineering/client/render/tile/SawmillRenderer.java
```diff
@@ -60,7 +60,7 @@ public void render(SawmillBlockEntity te, float partialTicks, PoseStack matrixSt
 			matrixStack.pushPose();
 			matrixStack.translate(1, .125, -.5);
 			float spin = te.animation_bladeRotation;
-			if(te.shouldRenderAsActive())
+			if(te.shouldRenderAsActive() && te.sawmillProcessQueue.size() > 0)
 				spin += 36f*partialTicks;
```
malte0811, line 64: This change looks like it won't work properly without changes to `SawmillBlockEntity`: This is only for "sub-tick" interpolation, the actual rotation is done in the BE. I'm also not sure I like the change in general, the rotating blade is a nice indicator that running into it will hurt.

malte0811 (changes requested):

src/main/java/blusunrize/immersiveengineering/client/render/tile/SawmillRenderer.java
voidsong-dragonfly: Good point! I forgot that, I'll put the interdiction there and stop the blade from hurting when it's not spinning

PR #5482, opened 2022-11-16
Use a different model for each crusher barrel, so they can intermesh

TeamSpen210:
This tweaks one of the crusher barrel models to offset the "blades", so they intermesh with each other instead of directly overlapping. 
image


BluSunrize:
Hrrm
Couldn't this be solved by mirroring the drum ingame? Do we require a second model for it?

TeamSpen210:
It can't, because the original drum has narrow sections on both ends, my additional variant instead has wide parts flush on either side (so one more in total). `-=-=-=-` versus `=-=-=-=`.

PR #5486, opened 2022-11-19
Machine Sound PR

voidsong-dragonfly:
Adds a bunch of new (and old!) sounds to machines, all under CC0
Machines with added sounds:
- Fermenter
- Refinery
- Assembler
- Blast Furnace Preheaters
- Mixer
- Bottling Machine
- Sawmill

Machines with old sounds added:
- Coke Oven (uses MC fire crackle sound)


Fluffton:
ZAMN!!! i can't wait to immerse myself EVEN MORE with these new sounds!!!

BluSunrize:
Sounds Good To Me!

A few minor things:

- Coke Oven does not loop well at all. There is a noticable "end" to the sound effect, and then it restarts.
- Assembler does not produce any sound in my testing
- The bottling machine sounds very loud and quite jarring (the pun is noted but unintentional)
- The idle sound of the sawmill also has a very obvious loop point. It sounds like it's winding up and down a bunch

Fluffton:
Honestly my only suggestion to this is diesel generator startup and stopping sounds. I absolutely hate the generator just creating noise out of thin air without any buildup. Other than that, this is cool!

PR #5523, opened 2023-01-07
New multiblock abstraction/API

malte0811:
This allows addons to use IEs multiblock BE implementations without using non-API code (or IE exposing too many implementation details). Additionally it abstracts away the master/dummy BE split, most "user code" does not have to worry about which BE it is running on any more.

Known issues:
- Currently there is no support for variable-sized multiblocks. It should be possible to add this later on without too much effort, but I'm not sure what the API should look like there.
- The documentation is very sparse/nonexistant. I'll try to add more documentation when I have some spare time, maybe in a week or so.
- Migration from old worlds may loose items/fluids/etc stored in multiblocks since the NBT format for the individual MBs changed. The general multiblock structure should be migrated correctly though.
  - Note: I've only tested migration on structure files so far, I'll try to test a full world some time next week.

I'm not expecting anyone to review the full +15k/-12k LoC diff. The "core" of the changes is in
- `blusunrize.immersiveengineering.api.multiblocks.blocks` (API) and
- `blusunrize.immersiveengineering.common.blocks.multiblocks.blockimpl` (implementation details).

If someone can review that and one or two "random" multiblock logic implementations it should be enough.


src/main/java/blusunrize/immersiveengineering/common/blocks/multiblocks/component/RedstoneControl.java
```diff
@@ -0,0 +1,140 @@
...
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.InteractionResult;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.phys.BlockHitResult;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class RedstoneControl<S> implements IMultiblockComponent<RSState>, StateWrapper<S, RSState>
```
SkySom: Is this perhaps a component that should be accessible from API?

src/api/java/blusunrize/immersiveengineering/api/multiblocks/blocks/util/CapabilityPosition.java
```diff
@@ -0,0 +1,34 @@
...
+
+	public static CapabilityPosition opposing(MultiblockFace multiblockPos)
+	{
+		return new CapabilityPosition(
+				multiblockPos.face().offsetRelative(multiblockPos.posInMultiblock(), 1),
+				multiblockPos.face().getOpposite()
+		);
+	}
+
+	public boolean equalsOrNullFace(CapabilityPosition other)
```
SkySom, line 30: What is the use of this method?

SkySom (commented):
Probably about half through, but did note two things.

src/api/java/blusunrize/immersiveengineering/api/multiblocks/blocks/util/CapabilityPosition.java
malte0811: We have a lot of places with code like
```java
if (cap == SOME_CAP && (side == null || (pos.equals(RIGHT_POS) && side == RIGHT_SIDE))
```
e.g. for energy capabilities that are exposed everywhere on the `null` side for TOP/Jade. That method is supposed to make this a little more readable, not sure how much that succeeded.

SkySom: Was more confused, why if side was null, you suddenly didn't care about position, but TOP/JADE makes sense

src/api/java/blusunrize/immersiveengineering/api/multiblocks/blocks/logic/IMultiblockLogic.java
```diff
@@ -0,0 +1,34 @@
...
+import net.minecraft.world.phys.shapes.CollisionContext;
+import net.minecraft.world.phys.shapes.VoxelShape;
+
+import java.util.function.Function;
+
+public interface IMultiblockLogic<State extends IMultiblockState> extends IMultiblockComponent<State>
+{
+	State createInitialState(IInitialMultiblockContext<State> capabilitySource);
+
+	// TODO split into collision and selection?
```
BluSunrize: I feel like we already have separate collision and selection on some multiblocks? So I think we do need this...

src/main/java/blusunrize/immersiveengineering/common/CommonProxy.java
```diff
@@ -6,6 +6,14 @@
...
  */
 
+/*
+ * BluSunrize
+ * Copyright (c) 2023
+ *
+ * This code is licensed under "Blu's License of Common Sense"
+ * Details can be found in the license file in the root folder of this project
+ */
+
```
BluSunrize: Why are there so many copyright statements in this file? O_o

BluSunrize (approved):
Tentative LGTM.

There is a *lot* of code to go through here. I presume if there are issues, they'll arise in use testing.

src/main/java/blusunrize/immersiveengineering/common/CommonProxy.java
malte0811: WTF did IntelliJ think it's doing… Apparently if you don't enable "update existing copyrights", it just keeps adding new copyright statements if at least one old one exists? I'll have to check if any other files are affected by this.

Edit: No, apparently just this file?? Maybe because it got two 2017 copyright statements in 495e2c5365235881285909cff65e087f945fdc1d?

PR #5532, opened 2023-01-17
Fix for a crash when Secondaries of a recipe have different item from main recipe result / Resolves #5506

### LINKED ISSUE #5506: Usupported operation thrown at `api/crafting/ArcFurnaceRecipe.java#L110`
### Minecraft Version

1.19.2*

9.0.0-153

### How did it happen?

https://github.com/BluSunrize/ImmersiveEngineering/blob/1.19/src/api/java/blusunrize/immersiveengineering/api/crafting/ArcFurnaceRecipe.java#L110
`NonNullList#add(object)` throws an unsupported operation exception and causes a crash, only `NonNullList#add(integer,object)` is valid.

Issue was found while trying to add custom recipes to Arc Furnace with Secondary output. 
Using field `"output": { "item": "example:id" }` instead of `"output": { "tag": "example:tag" }` causes the crash.

Testing Recipe used to trigger an exception:
```json
{
    "type": "immersiveengineering:arc_furnace",
    "additives": [],
    "energy": 51200,
    "input": { "item": "minecraft:raw_iron" },
    "results": [
        { 
            "base_ingredient": { 
                "item": "minecraft:iron_ingot"
            },
            "count": 2 
        }
    ],
    "secondaries": [
        {
            "chance": 0.75,
            "output": { "item": "minecraft:gold_ingot" }
        }
    ],
    "time": 60
}
```

### Crash Log

https://gnomebot.dev/paste/1054462962846421083

### This issue does not fullly explain the problem!

To get more details why this is happening, check out #5532

Kanzaji:
Better explanation of the issue: A NonNullList is not a dynamic list implementation. Its size cannot be changed after creation. If the secondary field could not be merged with the main recipe result, it is supposed to be added to the output, which is not possible due to the nature of NonNullList and caused a crash.
None of the IE recipes trigger this problem, but custom recipes from people who want to use secondary outputs different from the main output do.

In #5506, I misunderstood the actual problem, as using Item instead of Tag should work fine, and the real problem is described above.

The fix increases the size of the initial list from just the output size to the output size plus the size of all secondaries, even if they can be merged.
This results in some empty ItemStacks in the final list but fixes the crash caused by the secondaries being different from the main recipe result. Empty ItemStacks have no effect because the proper treatment of them is already in place, as far as I can tell.

### DIFF
--- a/src/api/java/blusunrize/immersiveengineering/api/crafting/ArcFurnaceRecipe.java
+++ b/src/api/java/blusunrize/immersiveengineering/api/crafting/ArcFurnaceRecipe.java
@@ -91,7 +91,8 @@ public class ArcFurnaceRecipe extends MultiblockRecipe
 	{
 		Random random = new Random(seed);
 		var output = this.output.get();
-		NonNullList<ItemStack> actualOutput = NonNullList.withSize(output.size(), ItemStack.EMPTY);
+		int remainingIndex = output.size();
+		NonNullList<ItemStack> actualOutput = NonNullList.withSize(output.size() + secondaryOutputs.size() , ItemStack.EMPTY);
 		for(int i = 0; i < output.size(); ++i)
 			actualOutput.set(i, output.get(i).copy());
 		for(StackWithChance secondary : secondaryOutputs)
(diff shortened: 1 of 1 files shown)


BluSunrize:
Please submit a pullrequest targeting the 1.18.2 branch, since that is our main development branch for now.

I attempted to change the target branch of this PR, but that is not viable, because it results in a massive diff.

malte0811:
Manually cherry-picked on 1.18.2 as 891e940909ab1e45a397057d4c54348f1f485921

PR #5544, opened 2023-01-25
Fix for Hemp Seeds dropping while using Modded Shears

Kanzaji:
### Issue Description
While using modded shears, Hemp Seeds still drop, because of the way loot modifiers are made 😅

### Fix Description
Change predicate from Items to Tag, to allow modded Shears to work as intended 😄


BluSunrize:
Thank you!

PR #5559, opened 2023-02-04
[DE Translation]: small fix for 1.18.2

TheScienceDemon:


BluSunrize:
This isn't even an error.
"Kreosot" is a valid German translation for "creosote", and anyone in the Ruhr Valley will know what you mean by "Koks" without needing to specify coal...

PR #5599, opened 2023-03-05
Fixed Improved Blast Furnace output capability facing

xieve:
The inventory handlers of the Improved Blast Furnace were previously pointed inwards, which meant that it was not usable with certain modded item transportation means that did not implement a container, like the Create mod's funnels. This has been fixed.

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/stone/BlastFurnaceAdvancedBlockEntity.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/stone/BlastFurnaceAdvancedBlockEntity.java
@@ -217,9 +217,9 @@ public class BlastFurnaceAdvancedBlockEntity extends BlastFurnaceBlockEntity<Bla
 		{
 			if(inputOffset.equals(posInMultiblock)&&facing==Direction.UP)
 				return inputHandler.getAndCast();
-			if(outputOffset.equals(posInMultiblock)&&facing==getFacing())
+			if(outputOffset.equals(posInMultiblock)&&facing==getFacing().getOpposite())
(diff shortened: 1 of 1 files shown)


BluSunrize:
I think our code is okay as is?
An Improved Blast Furnace's `getFacing()` returns the Direction it is pointing at. Here is an example in a test world:
grafik

That one has a facing of "east" and I am look towards "west" here, meaning I am looking at the eastern face of the block.
Which means, the output for items has to be facing towards "east" as well to extract the steel -> `facing==getFacing()`

I don't think the switcharound you are suggesting in this PR is correct...

xieve:
I'm not sure why exactly, but I have tested the code I've submitted in this PR, and it works. Here's the setup:
grafik
I'm running upstream IE right now, that's why the funnel is not outputting any steel. When I compile my fork, this exact setup does output steel via the funnel. Maybe the entire model is backwards? I don't know.

xieve:
Interestingly, the multiblock model is actually backwards in the manual.

xieve:
Hey there, could you please re-visit this PR? The testing is very simple, you can try it with a Create funnel or any kind of item pipe that does not implement a container. My change fixes the aforementioned problem on my machine. Let me know if there is any other information I can provide.

BluSunrize:
I did test this again, specifically with Create funnels, and I found that it does in fact break.

However, the issue is that `slagOutputOffset` and `outputOffset` were configured wrong. Your fix does in fact *work*, but then results in slag coming out of the front and steel coming out of the back.

So I am going to fix this by actually adjusting the two offset references.
Thank you so much for your assistance in debugging this though, and giving me the hint to try it with Create Funnel =)

PR #5604, opened 2023-03-07
Add sawmill recipes for Treated Wood

TeamSpen210:
Noticed that the 3 treated wood plank/stair/slab blocks couldn't be processed in the sawmill the same way other planks can, so added recipes for them.


BluSunrize:
I picked this one manually to avoid the conflict:

https://github.com/BluSunrize/ImmersiveEngineering/commit/3ffb58d6dcb03c8ba9e46043ed617c31772b6794

Thank you =)

PR #5676, opened 2023-05-06
Fix broken Arc Furnace output slots

StevenDoesStuffs:
This bug also exists in the other versions--I can PR that too if you'd like.


BluSunrize:
We aren't maintaining 1.16.5 anymore, but I will include this fix in 1.19

PR #5796, opened 2023-11-01
HempBlock: Update neighbours when growing top block

jrtc27:
Vanilla multi-block crops like SugarCaneBlock use setBlockAndUpdate when
placing the new block (equivalent to calling setBlock with update flags
3, i.e. UPDATE_NEIGHBORS | UPDATE_CLIENTS) rather than a direct call to
setBlock with flags 2 (i.e. just UPDATE_CLIENTS). However, HempBlock
currently just does the latter. Whilst vanilla's observer seems able to
pick up this update, AE2's Annihilation Plane cannot, and so will not
harvest the crop unless refreshed for some other reason (e.g. the
network is rebooted). Thus, more closely align the implementation with
blocks like SugarCaneBlock in order to play nicely with whatever trigger
mechanism AE2 is using.


BluSunrize:
Thanks! =)

PR #5798, opened 2023-11-03
a config flag to disable rendering blueprints

SirWalkerCZ:
Hi, I was having an issue where the game would crash upon putting a blueprint into the engineers workbench.
The issue only happens when you run IE alongside some kind of performance mods (Optifine, Sodium clones etc.).
I know you said you dont oficially support them, but this is such a simple fix that could stop some people from bothering you about it.
Only thing it realy does is add a flag to the client config that you can enable to stop from rendering the blueprint lines which seem to crash it whenever someone has shaders turned on
Here is an example crash log:
[latest.log](https://github.com/BluSunrize/ImmersiveEngineering/files/13252352/latest.log)
I kinda need this pulled so I can upload my modpack to curseforge so my less technically savvy friends can play it too...

Btw thanks for making this supercool mod :)


BluSunrize:
Sorry, we're not planning any further updates for 1.18, and mods like Optifine and Sodium fall under no support anyhow.

PR #5812, opened 2023-11-18
Improve jade tooltips

HermitOwO:
Show the inventory of multiblocks in the tooltips when looking at any part of the multiblock (previously the inventory would only show if the block has item handler capability)
Show the tanks of multiblocks in the tooltips when looking at any part of the multiblock (previously they don't show at all)


BluSunrize:
Thank you!

PR #5824, opened 2023-12-14
Fix External Heater breaking when used with an extra-fast cooking recipe.

TeamSpen210:
Though it's not used in vanilla, it's possible to adjust the `cookingTime` for furnace recipes. The heater's speedup behaviour didn't account for this, allowing it to increment the count beyond the expected maximum, breaking the furnace entirely since it uses `==` to check for completion. I altered the code to get the actual cooking time for the recipe and use that instead. While investigating, I also noticed that the furnace has a `RecipeManager.CachedCheck` instance used to avoid scanning the whole recipe list every tick, so I swapped the heater handler to use that. Haven't done timings, but I assume that might improve perf?

To test, I used the following recipe. Would it be a good idea to include that in the `test` package, maybe make a gametest? 
```json
{
  "type": "minecraft:smelting",
  "ingredient": {
    "item": "minecraft:dirt"
  },
  "result": "minecraft:diamond",
  "cookingtime": 25
}
```

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/util/VanillaFurnaceHeater.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/util/VanillaFurnaceHeater.java
@@ -25,24 +25,22 @@ public class VanillaFurnaceHeater implements IExternalHeatable
 		this.furnace = furnace;
 	}
 
-	int getCookTime()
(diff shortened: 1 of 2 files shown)


BluSunrize:
Thanks!

PR #5835, opened 2023-12-31
Add More Uses for Lead/Nickel

voidsong-dragonfly:
Adds more uses for nickel & lead in other recipes, as the metals have a tendency to pile up without very specific uses to drain them.

**NICKEL:**
Constantan is now used for radiator block 'piping': IRL copper-nickel alloys are used for corrosion resistance in condensers and radiators, and this provides another use of some nickel later in energy production.
image

**LEAD:**
Lead is used for red & white dye as lead oxide mix & lead carbonate, using oxidation for red and a simplified version of the Dutch process for white.
Red:
image
White:
image
Lead is also used for glass framing for tinted glass, reducing amethyst cost per block.
image


TeamSpen210:
Since the Dutch process for lead isn't immediately obvious, maybe it might be good to mention that in a comment, maybe also in the manual?

BluSunrize:
Yeah I have no idea how and why that white dye recipe is a thing.
Without any context, it feels extremely random.

Also I don't really see why one recipe uses lead nuggets and slag glass, and the other uses wire and plain glass. What's the reasoning here?

voidsong-dragonfly:
> Yeah I have no idea how and why that white dye recipe is a thing.
Without any context, it feels extremely random.

That's a very good point. I can see about adding that. The basic idea is "vinegar + rotting material + lead + time == lead carbonate" which I simplified to "ethanol ages to vinegar + rotten flesh + lead + (no time)". I'll see about adding an 'Industrial Dyes' section to the manual to describe the various dyemaking recipes. 

> Also I don't really see why one recipe uses lead nuggets and slag glass, and the other uses wire and plain glass. What's the reasoning here?

The idea was that using slag glass - which can be sturdier than normal glass, if more opaque - lets the user cut down on other materials (namely lead) in the glass. It's a way to use less lead in the new recipe but use slag glass. Wires is because it's a good half-ingot value that could be formed into the framing of the glass, nuggets because they're smaller than wires. 
** **
I have no particular attachment to the exact recipes in this PR, I should have specified that. I'll happily remove one or two here or there to make things clearer, and if you want one of the glass recipes one or changed, feel free to tell me.

PR #5851, opened 2024-01-19
Update ja_jp.json

TwelveYO21:
Sorted as well as the line in en_us.json


karakufire:
Your this pull-request looks me draft or work in progress.
I'll evaluate higher the parts that tries to fix the differences between the 1.19.2 and 1.20.1 than other changes.
But, I couldn't find the meanings following changes:
 + > Sorted as well as the line in en_us.json
   + I think there is no need to bother changing entries order.
 + Almost your addtions is taken by untranslated strings.
 + Lines at 175-184, what is the meanings of re-adding entries that remain English?
 + Lines at 774-783, they are **REAL** used. Read the manual and check shader list page.

Finally, if this pull-request is a work in progress, it should be draft. And if you pull-requested this as completed, there is no point in merging this request.

TwelveYO21:
Sorry for the late reply.
I just adapted it to en_us.json in 1.20.1. I can't help with the translation work, but I just made it easier to edit later.

karakufire:
At least the removes Lines at 774-783, `desc.immersiveengineering.info.shader.reference.*`, are not acceptable. Because they are still used in manual. 
If you still would like ImmersiveEngineering to merge this pull-request, you should revert this changes.

PR #5863, opened 2024-02-04
Fix earmuff behavior for long-running sounds

malte0811:
As discussed on Disord starting [here](https://discord.com/channels/270295681829306369/539145317547900938/1203665420645568532): The previous implementation only applied attenuation when starting the sound, so long-running sounds did not update when earmuffs were equipped/unequipped. The new approach uses a Mixin, but can handle such sounds correctly.

Note that this PR is on 1.20.4 since for a compatible implementation we need the injectors from MixinExtras. This library is provided by NF on 1.20.4, but not on 1.20.1.


BluSunrize (approved):
Pretty clean use of Mixins, I'd say.
Could be interesting to see what happens when multiple mods do this, haha.

BluSunrize:
Followup question: Is this something we should have considered for 1.20.1?

malte0811:
> Could be interesting to see what happens when multiple mods do this

It will "just work" with multiple mods, and that's essentially why it *isn't* on 20.1: this clean and compatible implementation only works with MixinExtras, which isn't provided by NeoForge on 20.1. I can try to put something together, but it will probably be much more complex.

PR #5870, opened 2024-02-12
Fix free energy from diesel generators filled with water (#5844)

### LINKED ISSUE #5844: Diesel generator in ATM9 makes energy from any fluid and does not consumes them (1.20; ATM9)
### Minecraft Version

1.20.1

10.0.0

### Expected Behavior

Diesel genarator, if wrong fluid is inserted, then no energy is generated, or it does not even accet the wrong fluid

### Actual Behavior

Diesel generator, if not one of the following fuidls is inserted into: creosote/ creosot oil from railcraft, gergtech, thermal or IE; biodisel frompneumaticraft or IE, does not consume the inserted fluid, BUT still generates power. If correct fluid is used, animations are played, if wrong fluid is inserted, there are no animations. Tryd with multiple fluids ex.: water from minecraft, fuels from other mods like thermal or gergtech, gotohell, milk or even the other fluids from IE like ethanol or concret. The only ones that worked were some sort of fluid dictionary equivalent of IE fluid, so ex.: pneumaticrafts biodiesel, but not industrial forgoing biodiesel.

### Steps to reproduce

1. Build valid diesel generator and tank, connect the two with IE piping, aplly redstone signal to the tank so it auto outputs.
2. Put any bucket or any amount of fluid into the tank, or inser the fluid in an other way into the generator
3. If not fluid I listed as working inserted, the generator does not play animations, does not consume any of the fluid, BUT still generates energy indifenetly

### Debug Crash Log

https://pastebin.com/wuAgXd1w

the-pandoran:
Moves the check that the generator's contained fluid is a valid fuel so it occurs before energy is generated. Fixes #5844

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/multiblocks/logic/DieselGeneratorLogic.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/multiblocks/logic/DieselGeneratorLogic.java
@@ -78,14 +78,15 @@ public class DieselGeneratorLogic
 					.map(CapabilityReference::getNullable)
 					.filter(Objects::nonNull)
 					.collect(Collectors.toList());
-			if(!presentOutputs.isEmpty()&&EnergyHelper.distributeFlux(presentOutputs, output, false) < output)
-				state.consumeTick--;
-			if(state.consumeTick <= 0) //Consume 10*tick-amount every 10ticks to allow for 1/10th mB amounts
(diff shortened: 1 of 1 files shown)


AWAKEN-1:
OMG this is what I was looking for, admins please merge it in.o(╥﹏╥)o

BluSunrize:
Thanks =)

PR #5893, opened 2024-03-18
Cloche Fertilizer Changes

voidsong-dragonfly:
Adds in new fertilizers for the cloche to balance the enormous amount of sulfur that comes from the Excavator.

Sulfur is the lowest fertilizer, at 1.1x
Nitrate is 1.2x, partly as it is infinite but needs cobblestone to be gotten.
Slag and Bonemeal are each 1.3x, partly because they take some effort to get.
_Industrial Fertilizer_, the IE option, is 1.6x. It needs a water bucket, sulfur, nitrate, and slag to craft. If you disagree with the recipe please do tell me, I just thought the water bucket felt right there. (And in IE, you'll need an assembler anyway)

The Cloche GUI got an update (the progress bar actually renders) and JEI categories got updated: the cloche one now displays more information in the GUI style, and there is a display for fertilizer growth modifiers. Both are shown below.

image
image


BluSunrize:
Thank you!

Issue #5914, opened 2024-04-15
Redstone Interface Connector not working with CC: Tweaked in 1.20.4


Sharparam:
### Minecraft Version

1.20.4

### Immersive Engineering Version

11.1.0-172

### Expected Behavior

Should be able to set the redstone values on the interface connector, just like in 1.20.1.		

### Actual Behavior

Nothing gets set on the interface connector.

### Steps to reproduce

1. Place a CC: Tweaked computer
2. Place a Redstone Interface Connector on top
3. Place a redstone lamp next to the computer
4. Place a Redstone Wire Connector on top of the lamp
5. Connect the interface to the wire connector with Redstone Wire Coil
6. Configure the connector on top of the lamp to be "output - red"
7. Set the top redstone output on the computer: `rs set top red true`

### Debug Crash Log

https://gist.github.com/Sharparam/b45c9678a0e9f2d258e73b595fdebbd9

BluSunrize:
@malte0811 CC:Tweaked Compat seems to have been fully removed.

As far as I can see, it should be able to restore it, but the changes to capability registration and our own Multiblock logic necessitate some changes to everything.
I'm assigning this one to you, but it's low-prio, don't stress yourself.

malte0811:
Fixed by 2952ec85343c3ad53b15eae764c16f442776f4b8 and 98320a36858c8e21e5e23f8ac0e573bd6279a79b

PR #5925, opened 2024-05-01
Update CC:T integration to 1.20.1

SquidDev:
This is a couple of fixes to update ComputerCraft/CC: Tweaked to match the 1.20.1 API.

 - Remove `TaskCallback`, as that is now exposed in the public API.
 - Fix arguments not being marked as escaping when running off-thread.
 - Replace `ItemStack`/`FluidStack` conversion with CC's detail providers. This should be backwards compatible (the fields exposed appear to match), but happy to revert this if preferred!


BluSunrize:
Sorry, we only do bugfixing on 1.20.1

New features are 1.20.4 only. If you would like to, you update to that version, however?

SquidDev:
Ahh sorry, wasn't sure so just targetted the default branch! Yep, will update the branch and retarget the PR.

SquidDev:
I've pushed something which partially works, with the exception that the code doesn't subscribe to the `RegisterCapabilitiesEvent` event. It's a bit tricky for compat modules to subscribe to this event for a couple of reasons:

 1. The event is fired during the registry stage, so before compat modules are loaded. And we can't move module loading any earlier, as the config values haven't been loaded yet.
 2. The mod event bus isn't accessible to compat modules. We could definitely pass it around as an argument, but didn't feel worth making this change without addressing 1).

Both are definitely fixable[^1], but I didn't really feel comfortable making any changes without further discussion. 

[^1]: I tested by switching CC to be an `EarlyIECompatModule`, but that's probably not a good solution.

BluSunrize:
...yeah this is why I assigned @malte0811 to the original issue, because I don't know how to get those capabilities in there anymore since the old event has stopped existing.

SquidDev:
Ahh sorry, I hadn't spotted #5914! Definitely agreed that this isn't high priority, so happy if you want to close this and use it as a reference for a future fix :).

SquidDev:
Thank you both!

PR #5955, opened 2024-06-17
Diesel tool sounds - First Draft / Feedback desired

### LINKED ISSUE #3262: [Suggestion] Sounds for Mining Drill & more Machine Sounds
Hi, absolutly love the mod, amazing work and many thanks!
Something that bugs me a bit is the lack of some immersive sounds,
especially when mining with the drill, it's so quiet. 
It's a powerful tool and I think adding some motor sounds to it while active would really enhance the 
feeling using it. A config option to disable or tweak the volume of it could be added for people who don't like it.
Same goes for some other machines. 
Some hydraulic and motor sounds for most machines would do so much for the overall immersion.
It's an industrial mod and it should sound like one, I think.
Sounds are an often overlooked feature, but they do so so much for the overall immersion I believe.
What do you think ?
Thank you for this wonderful mod and all your hard work!

MalkContent:
This works well enough, but it's like 13 workarounds in a trench coat.
I have some thoughts about how to do some things differently, but I figured a) perfect is the enemy of good and b) it's probably a good idea to get some feedback on this to see if just what the PR does is in the right vein
technical feedback of course welcome, too ^^

resolves https://github.com/BluSunrize/ImmersiveEngineering/issues/3262


MalkContent:
resolves https://github.com/BluSunrize/ImmersiveEngineering/issues/3262

src/main/java/blusunrize/immersiveengineering/common/util/sound/DieselToolSoundHandler.java
voidsong-dragonfly, line 1: ~~Why have all of this? Why not do something similar to how `handleTileSound(...)` in `ClientProxy.java` works? That seems like it would both be easier and more effective, though I may be off my rocket and confused here.~~ Once talking with Malk, I think this is probably not reasonable to do.

voidsong-dragonfly (changes requested):
First thing, Malk: These sounds are extremely loud and close to the player. I would suggest making the sounds maybe a quarter again as quiet to ensure they're not too disruptive to the player experience.
Second thing: The harvest detection you implemented for the drill/busszaw does not work particularly well **between blocks of the same type**. When breaking grass blocks just as a test, the drill seemed to keep puttering up and down and not actually hitting a stride and running. Within one block it worked great.
Third thing: All the drill sounds and the saw sounds except for the saw cutting sound great. However, the saw cutting sounds... very electronic? And might be better for the angle grinder only, and have something else for the sawblade. Maybe the sawmill?

MalkContent:
to First: I put in a (hopefully temporary) static for you to control all the sounds volume with. I recon it's easier for you guys to adjust the volume until you find something you're comfortable with, then I can adjust the sounds themselves to that level. My original take was "loud noisy tool, use ear defenders when annoyed"

to Second: after having cleared up what you mean (sound jumping up and down between quickly digged blocks), I added a fade sound after digging, which does sound much nicer imo. I had already mulled over doing that previously, but originally decided against it, because I wanted to KISS it to not make Blu groan harder than I assume this PR does anyways x)

to Third: I agree about the saw. Had a hard time to find a good sound that does not sound "far away with echo, in a factory hall". I hanged it so the sounds depend on the ItemStack, not just the Item when fetched, so now the rockcutter and grinding disk have the grindy sound and the good old saw blade has a copy of the saw mill's saw_full.ogg as a temporary place holder.

-> Experimenting, because I initially did not understand your description in "Second" correctly made me find a bug with a double sound when you start digging under specific circumstances, though. Have not nailed it down yet, though.
And for the saw (blade) sawing sound, I'll have another look around, but that'll probably be a bit later in the weekend.

MalkContent:
fixed the bug ^_^
now just gotta see about the sawing sound

voidsong-dragonfly:
There's still a bit of the "jumpiness" I mentioned, but you're right that there is a lot less of it and it sounds less annoying. Not sure if there is a good way to fix it, with how the delays between block breaking works. For now, this seems... fine enough? I'll wait for Blu to weigh in on what she thinks about the PR, though.

MalkContent:
yea, I can understand that. I'm open to ideas there, but my ideas on "how to do this differently" basically boil down to "extend the digging sound beyond the point where the block breaks to bridge the gap", which probably sounds weird with the block breaking, but the digging sound still going, or "do button press hitscan stuff and check if the player <would> continue digging, if he keeps the button pressed", which could maybe work nicely, but I also am not looking forward to working that into the state machine.

And, jumpiness aside, my argument in favor of the current behaviour is "it does tell you clearly when you're digging and when you're not".

Maybe the sounds themselves could be addressed to smoothe the jumpiness somewhat. But I'm taking a couple days to cleansy my ears' palate from listening to saw sounds for hours x)

MalkContent:
hoh boy. I'll look at that datagen error with fresh eyes some later date

src/main/java/blusunrize/immersiveengineering/common/items/BuzzsawItem.java
```diff
@@ -20,6 +20,8 @@
 import blusunrize.immersiveengineering.common.register.IEDataComponents;
 import blusunrize.immersiveengineering.common.register.IEItems.Misc;
 import blusunrize.immersiveengineering.common.register.IEItems.Tools;
+import blusunrize.immersiveengineering.common.util.IESounds;
+//import blusunrize.immersiveengineering.common.util.ItemNBTHelper;
```
BluSunrize: We do not do commenting out of imports.

src/main/java/blusunrize/immersiveengineering/common/util/sound/NoisyToolSoundGroup.java
```diff
@@ -0,0 +1,292 @@
...
+				play(new NoisyToolMotorSoundLooping(noisyToolItem.getBusySound(noisyToolStack).value(), newMotorState));
+				break;
+			case FADING:
+				play(new NoisyToolMotorSoundFinite(noisyToolItem.getFadingSound(noisyToolStack).value(), newMotorState, FADE_DURATION));
+				break;
+			case ATTACK:
+				if(propagate)
+					updateHarvestState(null, false); // todo: check if this is really necessary. Not that it hurts.
+				play(new NoisyToolMotorSoundFinite(noisyToolItem.getAttackSound(noisyToolStack).value(), newMotorState, ATTACK_DURATION));
+				break;
```
BluSunrize, line 142: Attack sounds are currently not playing for me?

src/main/resources/assets/immersiveengineering/lang_base/en_us.json
```diff
@@ -462,6 +462,13 @@
   "subtitle.immersiveengineering.chargeSlow": "Charging capacitors",
   "subtitle.immersiveengineering.spark": "Electrical sparks",
   "subtitle.immersiveengineering.railgunFire": "Railgun fire",
+  "subtitle.immersiveengineering.drill_motor": "Drill motor running",
+  "subtitle.immersiveengineering.drill_attack": "Drill revs up high",
+  "subtitle.immersiveengineering.drill_harvest": "Drilling",
+  "subtitle.immersiveengineering.buzzsaw_motor": "Buzzsaw motor running",
+  "subtitle.immersiveengineering.buzzsaw_attack": "Buzzsaw revs up high",
+  "subtitle.immersiveengineering.buzzsaw_harvest_sawing": "Sawing",
+  "subtitle.immersiveengineering.buzzsaw_harvest_grinding": "Grinding",
```
BluSunrize, line 485: The buzzsaw harvest subtitles do not match what it listed in the sound file

BluSunrize (changes requested):
I finally checked this over, and I actually quite like it!

I've noticed two things that should be addressed:
1. The sound file lists `buzzsaw_sawing` and `buzzsaw_grinding`, but the lang file entries are `buzzsaw_harvest_sawing` and `buzzsaw_harvest_grinding`
2. The sound events for attacking seem to not actually play?

BluSunrize (commented):
LGTM! Thank you so much <3

BluSunrize (approved):
LGTM! Thank you so much <3

PR #5967, opened 2024-06-22
Add in Opening for Scaffold Covered Ladders

### LINKED ISSUE #2710: [Feature request] improvements for the new covered ladders
I tested the new ladders (the aluminum version). They work great but there are some minor details:
The metal ladder can't be placed at the side of a silo or tank - but the covered one can. This means you can't use them to climb up to a silo as you can't enter the ladder.

So it would be nice to be able to place metal ladders at the side of a tank/silo - or have the option to open the front of a covered ladder using the hammer.

voidsong-dragonfly:
Fixes #2710.

Scaffold Covered Ladders now have a new feature: shift rightclick with the hammer to 'open' them, removing the front face and allowing the player to walk into them! This removes the need for hanging ladders below scaffold ladders where they may not otherwise be supported.

Additional content: Comes with a necessary new interface for the hammer, `IHammerBlockInteraction` for implementation by blocks that want to interact with hammer shenanigans themselves, rather than have their BEs do it.


TeamSpen210:
Nitpick: naming the blockstate property `active` is a bit odd, would be better to make it something like `opening`.

BluSunrize:
> Nitpick: naming the blockstate property `active` is a bit odd, would be better to make it something like `opening`.

"open" I think would work well.
Can you adjust that, void?

voidsong-dragonfly:
Oh, I had it like that because IE already had the active property as a boolean! Yeah sure I'll add an `open` property.

PR #5972, opened 2024-06-23
Yet Another Language Update!~

AriaElidove:
Updated and Organizes the `en_us.json` file to be more consumable to a reader, plus some other minor fixes!

Also added language entries IE fluids because Jade was very unhappy without them x3


BluSunrize:
Mmh, no, sorry.
I'm not taking a full reorganization of our language file as a PR-

PR #5973, opened 2024-06-23
Resolve #5922

### LINKED ISSUE #5922: [Suggestion] Make different types of scaffoldings recognizable by name
### Explain your idea

While I was building the Distillation Tower from Immersive Petroleum, I didn't notice the recipe required `block.immersiveengineering.slab_steel_scaffolding_grate_top` rather than the `block.immersiveengineering.slab_steel_scaffolding_standard` I was trying with, because the names of both blocks are the same (`Steel Scaffolding Slab`) and the thumbnails are very similar even with GUI size x3.
javaw_RcFCzOiEPM

Only after going through Immersive Petroleum's issues I found out that I need a different slab (which I didn't even know existed), which I can only differentiate from the inventory if I used advanced tooltips (F3 + H):

Source strings:
https://github.com/BluSunrize/ImmersiveEngineering/blob/ab6e863b48530c0ff2c7f1aca47a8cebedcde942/src/main/resources/assets/immersiveengineering/lang/en_us.json#L990-L1007

To solve this, we could:
- Rename the blocks ("Steel Scaffolding", "Grated Steel Scaffolding", and "Wooden Steel Scaffolding"?)
- Update the textures to make the standard steel scaffolding and the grated steel scaffolding more visibly different, and maybe even bring more awareness of the variants so players are aware of their existence.

*Note: I couldn't find a suitable template to report this issue, it's neither a bug nor it's a crash, just an annoyance.*

AriaElidove:
Standard Scaffolding is as is, Scaffolding Grates are "Grated", and Wooden Scaffolding are "Treated"


BluSunrize:
This has failed checks due to the fact that data generation for the language file was not run.
I can fix that locally however.

PR #5989, opened 2024-06-30
Improve Utility of Gunpowder Barrel

### LINKED ISSUE #3692: [1.14.4 | 0.14-93] [Suggestion] Gunpowder Barrel issues
TNT in vanilla minecraft has become more useful, it is now able to be dispensed from a dispenser in its ignited form and also returns 100% of the drops of the blocks it destroys.
Gunpowder Barrels (GB) could be made to mimic that behaviour.

Explosions from GBs currently also do not support sympathetic explosions of other GBs or TNT. TNT explosions however do ignite GBs.

Finally, in order to set GBs apart from TNT in a way to not just make it a reskin with different ressource costs, GB explosions could be made to not destroy already dropped items. I have no strong oppinion on what it should do here, just put it out as a thought.

### LINKED ISSUE #5904: Making the Gunpowder Barrel actually useful.
### Explain your idea

So, at the moment, the gunpowder barrel item is pretty useless, but it looks cool and goes with the theme. So, here's an idea of mine for how to make people actually want to use them: Taking some inspiration from TerraFirmaCraft's Powderkeg, it could have a UI where up to 9 gunpowder can be added, and a variable blast radius depending on how much powder there is. It would be more efficient with its gunpowder use than TNT, with adding an equal amount of gunpowder pieces as are used to craft TNT resulting in a larger blast radius, however this would be offset by chain detonations not being possible, like it is currently. This would keep its original design intention as a mining explosive, better to use for opening up cave passages or easily harvesting large clusters of ore, but worse at blowing up structures.

### LINKED ISSUE #6068: Gunpowder Barrel Destroy Items
### Minecraft Version

1.20.4

11.6.1-181

### Expected Behavior

Use a Gunpowder Barrel to break cobblestone. The Barrel was surrounded with 26 stones.
Hope it can drop 26 cobblestone as the manual said.

### Actual Behavior

It only drop 11.

### Steps to reproduce

1. Surround a Gunpowder Barrel with 26 stones or other blocks.
2. Activate it.
3. Some of the blocks was destoryed, not dropping items.

### Debug Crash Log

<skipped>

voidsong-dragonfly:
The intent of this PR is to make the gunpowder barrel a useful tool for dealing with large volumes of rock.

As a part of this PR, gunpowder barrels are:
 - Instant explosions once again
 - Directional explosions that are biased against the direction with the most "hardness"
 - More powerful than TNT because they are actual bombs and not "gunpowder mixed with sand"
 - Better for antipersonnel damage when on the surface, but better at breaking rocks when inside rock

Fixes #5904, Fixes #3692, Fixes #6068

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/client/render/entity/IEExplosiveRenderer.java
+++ b/src/main/java/blusunrize/immersiveengineering/client/render/entity/IEExplosiveRenderer.java
@@ -8,10 +8,9 @@
 
 package blusunrize.immersiveengineering.client.render.entity;
 
-import blusunrize.immersiveengineering.common.entities.IEExplosiveEntity;
+import blusunrize.immersiveengineering.common.entities.GunpowderBarrelEntity;
 import com.mojang.blaze3d.vertex.PoseStack;
 import org.joml.Quaternionf;
-import org.joml.Vector3f;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.renderer.MultiBufferSource;
 import net.minecraft.client.renderer.block.BlockRenderDispatcher;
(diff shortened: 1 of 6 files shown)


voidsong-dragonfly:
This PR is still a slight bit wonky, specifically sounds do not always trigger for gunpowder barrels. However, it's at a point that I find _acceptable_ and I am tired of banging my head against a wall to implement a niche feature for mining.

Issue #6078, opened 2024-10-20
Game Crash with Sodium and Immersive Engineering when setting a garden cloche


zuken1997:
### Minecraft Version

1.21.1

### Immersive Engineering Version

ImmersiveEngineering-1.21.1-12.0.0-182

### How did it happen?

Hello i have a problem (1.21.1) with the new engineering version, I placed a garden cloche on my server and after that my game crashed. it looks like it's a bug with sodium, can you fix it please?

[crash-2024-10-20_22.53.01-client.txt](https://github.com/user-attachments/files/17452072/crash-2024-10-20_22.53.01-client.txt)

### Crash Log

https://gist.github.com/zuken1997/4adb17dfc4a6213a6ec302b62f2bd8dc

malte0811:
That needs to be fixed in Sodium.

jellysquid3:
I'd prefer to not disable critical optimizations in Sodium. It isn't clear to me why Immersive Engineering is re-implementing particle rendering in the way it is, since it will completely break draw batching and hurt performance badly, while also circumventing any optimizations that Sodium can apply.

Can you supply some insight [here](https://github.com/CaffeineMC/sodium-fabric/issues/2827)? It may be possible that Immersive Engineering doesn't need to use this VertexConsumer whatsoever but I don't understand your code fully.

PR #6079, opened 2024-10-21
Remove CustomParticleManager

IMS212:
This was used to fix a Vanilla bug where opaque particles could not show behind translucents. This has been fixed in NeoForge itself.

This has not been fully tested, however it seems to work in both cases it is used.


jellysquid3:
This should also reduce the number of draw calls for Immersive Engineering's block entities (since switching between particle types incurs a change in render layers, which flushes the current draw call.)  And it will enable better compatibility with Sodium while allowing our optimizations to continue being applied (see #6078.)

IlichiNiKoka:
is there any possibility to use the same way and method to fix this bug in 1.20.1 forge version ? Lotta thanks !
Edit: I'm sorry it's solved, use Embeddium instead :) sry .

PR #6087, opened 2024-10-25
Added About Section in README.md

Shubhrodip-Samaddar:
Added About Section in README.md #HacktoberFest.


BluSunrize:
I mean that's not even very accurate. Damien hasn't worked on this in years, Malte's contributions are entirely left out, and we never really set out to emulate AC either =p

Issue #6103, opened 2024-11-05
[Suggestion] Translation of hard-coded strings


momo-i2:
### Explain your idea

It's hard-coded in CircuitTableScreen.java at line 206-208.
![image](https://github.com/user-attachments/assets/b8117570-bf27-4b2c-afb7-7ef3d34246be)

PR #6104, opened 2024-11-05
Fix hard-coded strings

momo-i2:
This is just a Fix to make hard-coded strings translatable.

#6103


BluSunrize:
Why did you close this?
It would have been a very sensible improvement ^^

momo-i2:
Unfortunately, I have a quick temper and (I've had a lot of communication with them on Discord tho) no one has responded in over a month in here, so I decide that no one is willing to merge, I deleted the entire repository as a sorting out. The only thing to fix is to add 3 lines and change 3 lines, so feel free to do so. As for the translation, I am satisfied with using the resource pack as long as I can understand the content when playing with my own translation, so I will use the resource pack from now on. The previous translator seems to be doing translations of other industrial mods, so if you are lucky, previous translator may be able to do this one as well.

PR #6106, opened 2024-11-06
Add Japanese manual and Update Japanese translation

momo-i2:
Partially renamed (This translation don't include #6104)
If you think there is a mistranslation or a better translation than the current one, please report it in the comments. The manual has many parts that are difficult to read due to layout problems (there is room for improvement in the future).


momo-i2:
Fixed energy units name.

PR #6139, opened 2025-01-04
Add builtin compat with Repurposed Structures

TelepathicGrunt:
This will add to Repurposed Structures's villages automatically. The way it works is the special rs_ folders are read by Repurposed Structures and automatically does the piece injection. No code needed on your side.

Note, I setup the houses differently than what you currently do. I have 1 house for every village type. And then have an inner jigsaw that spawns a random workstation. This way I do not need multiple copies of the same house just to change the workstation and makes adding new workstations easier.

If you wish to have these files only apply when a config is set to true, you would simply grab the `repurposed_structures:json_conditions` registry on each loader and then add the `"condition"` field in the `rs_pool_additions` and `rs_pieces_spawn_counts` json files to point to the condition you registered. 

NeoForge: https://github.com/TwelveIterationMods/Waystones/blob/6a54aeeb05b88b24cf61722bd151e0362f2429ec/neoforge/src/main/java/net/blay09/mods/waystones/compat/RepurposedStructuresIntegration.java#L12

The JSON condition lines: 
https://github.com/TwelveIterationMods/Waystones/blob/6a54aeeb05b88b24cf61722bd151e0362f2429ec/common/src/main/resources/data/waystones/rs_pieces_spawn_counts/village_badlands.json#L8

https://github.com/TwelveIterationMods/Waystones/blob/6a54aeeb05b88b24cf61722bd151e0362f2429ec/common/src/main/resources/data/waystones/rs_pool_additions/villages/badlands/houses.json#L13

Datapack form of this PR if wish to test it outside PR:
[Repurposed_Structures-Immersive_Engineering_PR-Prototype.zip](https://github.com/user-attachments/files/18308675/Repurposed_Structures-Immersive_Engineering_PR-Prototype.zip)


TelepathicGrunt:
Also, let me know if you want me to add you to the 1.20.1 compat datapack project and give you a cut of cursepoints/modrinth revenue.
https://www.curseforge.com/minecraft/texture-packs/repurposed-structures-immersive-engineering

If so, let me know what % you want.

TelepathicGrunt:
@BluSunrize oh yeah one more thing, test this without my mod present for any errors in the console. If there are errors, then this might need to be an internal datapack that is loaded by event after a modid presence check

PR #6156, opened 2025-02-19
Fix #6149, arc furnace secondary chances flipped

### LINKED ISSUE #6149: Arc Furnace's secondary output chance is inverted
### Minecraft Version

1.21.1

Any version from 1.19 to 1.21, possibly earlier ones as well

### Expected Behavior

A secondary output in an arc furnace recipe should have the reported chance to be produced.
```
"secondaries": [
        {
        "chance": 0.99,
        "output": {
            "item": "minecraft:gold_ingot"
        }
        }
        ],
```
This should have a 99% chance of producing a gold ingot, which is also stated by JEI

### Actual Behavior

The output actually has a 1% chance to be produced.
It seems like the condition to roll a secondary output in ArcFurnaceRecipe should be a < rather than a >
```
if(secondary.chance() > random.nextFloat())
				continue;
```
This means that the chance parameter acts as the chance to NOT give that output.
It can technically be circumvented by setting the change to the opposite value (say like, 0.3 instead of 0.7) but then JEI would show misleading info.

### Steps to reproduce

1. Create an arc furnace recipe and load it with any datapack loading method.
2. Set the secondary output chance to something really high or really low
3. Test the recipe in game and see that the opposite result happens.

### Debug Crash Log

https://pastebin.com/3T29dtFZ

voidsong-dragonfly:


BluSunrize:
Okay fine, I can't base it against 1.20.4, but the issue definitely exists there too.
I'll be cherry picking this.

PR #6158, opened 2025-02-19
[1.20]Fix ArrayOutOfBoundsException Caused By Parallel Mod Loading

khjxiaogu:
Because of the parallel mod construct and loading, extension using api have a chance to load IE api classes at the same time as IE does, and there's a rare chance that the make method is called multithreaded, causing an ArrayOutOfBoundsException.


BluSunrize:
I'm not releasing for 1.20.1 anymore, but I've included this in 1.20.4.

Thank you! =)

PR #6204, opened 2025-04-02
Fix Diesel Generator burning fuel for too short of time, closes #6203

### LINKED ISSUE #6203: Diesel Generator can consume up to 8 times its fuel requirement
### Minecraft Version

1.21.1

12.1.0-185

### Expected Behavior

With a single refinery supplying a single generator with a full electrical load, the refinery's production of biodiesel should outpace the generator by 4x, filling up the fluid buffer in both the generator and refinery.

So, the generator should consume the expected amount of fuel once every 10-ticks (40mB every 10 ticks with biodiesel).

### Actual Behavior

Even though the refinery outpaces the generator, there is not an excess of biodiesel that fills the fluid buffers.

With a single refinery and generator, the generator will consume as much fuel as it can every tick, so long as its fluid buffer is less than 40mB.

So, with a single refinery, producing 16mB/tick of biodiesel, the generator will consume the 16mB in one tick, emptying its fluid buffer, but only run for that tick, consuming another 16mB in the next tick. (This also occurs with 2 refinery's at 32mB/tick)

If the generators internal fluid buffer contains less than 10000/burnTime of fuel (40mB in the case of biodiesel) then regardless of the amount burned, it only runs for 1-tick, and will try to consume more fuel in the following tick.

### Steps to reproduce

1. Setup a refinery and disable it with a redstone signal.
2. Setup a generator, ensuring that there is no biodiesel in the generators buffer, and the generator has never ran.
3. Fully load the generators electrical output (with creative accumulators set to input, excavator, etc.).
4. Fill the refinery with plant oil and ethanol, then enable the refinery.
5. Observe the generators fluid buffer, it will not fill up, but the refinery is producing more than the generator should need.

Notes: 
- If the generator has ran before with more than 40mB of biodiesel, there only seems to be a partial chance of this issue occurring.
- This was first noticed on a dedicated server, but reproduced in a singleplayer world.

### Debug Crash Log

https://gist.github.com/dthomas1202/37e85ef85391ee366b1536ed04914bb7

dthomas1202:
Fixes preemptive cast to int when calculating how long the generator should run on small amounts of fuel.

If the generator did not consume the full amount of `toConsume` then `state.consumeTick` would be truncated to 0, causing the generator to consume fuel again on the next tick.

Now the generator will calculate the correct amount of ticks to run for given a partial amount of fuel.

Fixes #6203

### DIFF
--- a/src/main/java/blusunrize/immersiveengineering/common/blocks/multiblocks/logic/DieselGeneratorLogic.java
+++ b/src/main/java/blusunrize/immersiveengineering/common/blocks/multiblocks/logic/DieselGeneratorLogic.java
@@ -96,7 +96,7 @@ public class DieselGeneratorLogic
 					{
 						if(!active)
 							active = true;
-						state.consumeTick = 10*(int)(fluidConsumed/toConsume);
+						state.consumeTick = (int)(10*(fluidConsumed/toConsume));
 					}
 					else if(active)
 						active = false;


BluSunrize:
Thank you!

PR #6244, opened 2025-05-20
Update es_es.json (1.20.1 - 10.2.0-183)

GGlangf:


BluSunrize:
We are not maintaining 1.20 anymore, please direct this pullrequest against the 1.21.1 branch instead.

Also, please include the changes from #6245 and #6246 in this PR, there is no reason to do three separate ones.

Issue #6253, opened 2025-05-27
Silo drops iron sheet metal


BluSunrize:
### Minecraft Version

1.21.1

### Immersive Engineering Version

Any

### Expected Behavior

It should drop what I built it with

### Actual Behavior

The first block broken is always iron

### Steps to reproduce

-

### Debug Crash Log

-

omerbenda:
Hey! dropped a [pull request](https://github.com/BluSunrize/ImmersiveEngineering/pull/6256#issue-3111792458) for this issue

PR #6256, opened 2025-06-02
Fixed multiblocks not dropping original block in the place mined

omerbenda:
[Original Issue](https://github.com/BluSunrize/ImmersiveEngineering/issues/6253#issue-3092993358)


BluSunrize:
Appreciated, but we already have a state cached during disassembly we can use, so I used that.

Fixed by https://github.com/BluSunrize/ImmersiveEngineering/commit/1f8167052ab38e968032fba68be0931a47b73c37

Issue #6251, opened 2025-05-26
HV Transformers dont distribute power correctly


CoconutMage:
### Minecraft Version

1.21.1

### Immersive Engineering Version

12.2.0-187

### Expected Behavior

Adding MV connectors to the output side of an HV transformer should increase power draw across the transformer by 1k a connector until hitting 8k and the wire melts.

### Actual Behavior

6 outputs = 6k power (expected), 7 outputs = 5k power, 8 outputs = 4k power
This does not occur on 1.20 version of mod
![Image](https://github.com/user-attachments/assets/ed9259f6-a3bb-4d57-93e1-010838b7fb57)
![Image](https://github.com/user-attachments/assets/a1e531d2-7cab-4090-a05b-c13a432f7559)
![Image](https://github.com/user-attachments/assets/7220d3b5-9fa6-42dd-985b-2e4f6ec3a8d6)

### Steps to reproduce

Mekanism energy cubes used for simplicity, but accumulators or any other generator on the left and machine on the right produces the same behaviour
![Image](https://github.com/user-attachments/assets/1529880e-8c7d-48e9-af96-67adcb73909c)

### Debug Crash Log

[debug.log head](<https://gnomebot.dev/paste/mclogs/PLKJ1DA>) / [tail](<https://gnomebot.dev/paste/mclogs/GAoDTsm>) (~19MB & ~157k lines)

PR #6298, opened 2025-07-24
Fix Transformer Limit Issues (#6251)

voidsong-dragonfly:
This fixes _several_ power transfer issues I introduced surrounding transformers when I introduced power throughput limiting to them in 1.20.4; it should fix the entire system as far as my tests have confirmed.

It also introduces better documentation to the power transfer code, so I can fix issues more quickly in the future.


BluSunrize:
Thank you!

PR #6335, opened 2025-10-14
Create centralized space to limit dye colors to only vanilla ones whe…

PssbleTrngle:
This PR fixes two issues that occur when playing with mods that add more dye colors:
- the changes in `Utils.java` allow these new dyes to color conveyors and balloons
- all other changes limit the dye colors used for signaling to the vanilla ones, even if others are present.

Especially the second change allows it to work with mods like dye depot without simply crashing on startup in many different ways.
What this PR does not fix and is in no way the responsibility of IE is missing data & assets for chutes and sheet metals.
I have still allowed them to be automatically created by looping through `DyeColors.values()` instead of only using vanilla colors here. This is in order for other mods, specifically [Dye The World](https://github.com/PssbleTrngle/DyeTheWorld) to add these assets and data and allow this compatibility

### DIFF
--- a/src/api/java/blusunrize/immersiveengineering/api/utils/SignalColors.java
+++ b/src/api/java/blusunrize/immersiveengineering/api/utils/SignalColors.java
@@ -1,0 +1,29 @@
+/*
+ * BluSunrize
+ * Copyright (c) 2025
+ *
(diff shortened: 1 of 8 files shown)


SammySemicolon:
Would love to see this merged

PR #6341, opened 2025-11-16
Fix GuiReactiveList Text-Scrolling being Too Fast

TwistedGate:
Regardless of FPS count the Speed now stays constant and is neither too fast nor too slow.

---
Basicly copied from https://github.com/TwistedGate/ImmersivePetroleum/commit/e0941ebb3357d0c92b0da67df207a4ac3e374687


TwistedGate:
Replacing with proper PR.

PR #6426, opened 2026-07-12
Lighting and major rendering performance fixes for multiblocks

tgstyle:
Fix lighting artifacts and duplicate rendering on split multiblocks

Large multiblocks built from split models (excavator, crusher, etc.) rendered their geometry as much as 7 times over. Fixed, along with a stale-cache bug that kept old geometry around after a resource reload.

Already proved working in Immersive Technology 1.20.1/1.21.1, and Immersive Fixes mixins for IE 1.20.1.


tgstyle:
Cancelling for now, I did discover an issue with multiblocks not respecting ambient lighting underground.

PR #6444, opened 2026-09-02
Add idle + disabled turret rotations

Raverbury:
This adds a small feature in the form of idle rotation as well as a disabled animation for turrets. The end result should look like the video attached.

https://github.com/user-attachments/assets/61cd0097-432b-49dc-a920-80466c1ea9db

I've tried to keep it simple but it still winds up being fairly long. It's understandable if you want to forgo this or use your own implementation.


PantherOS1991:
This is actually great.  This would be really cool.  Also... being able to destroy them would be really neat too (as a reference to half life and metroid.)

DereC4:
hi

